PR #2, opened 2018-01-31
John Spraul, RWD, FINISHED

Smithface:
creating pull request early so I don't forget later

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -269,6 +269,14 @@ header .logo {
     margin-top: 52px;
   }
 }
+@media (max-width: 300px) {
+  .cta {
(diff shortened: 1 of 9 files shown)


tsj7:
Looks great so far man. Well organized and easy to read. I like that you did commits after working the width of each section (laptop, tablet, etc.). That is a great habit to get in. Even if you are just committing locally and then do one or two pushes to github per day. If you want/need to revert to previous code, this will come in handy. Great job!

Smithface:
it gets a bit rough in between 400px and 300px widths (the circle image doesn't fit), but I'm okay with it.

PR #3, opened 2018-01-31
Daniel Abbott - responsive web design - I am clearly not a visually oriented designer.

Mephestys:


tsj7:
Nice work Daniel, your organization and structure looks great. A suggestion maybe get in the habit of committing a bit more often. Example: once laptop functionality is set. You don't have to push to github each time, but even committing locally after different bench marks can help organize in the case you need to come back to a code block. Looks really good though, keep it up man!

Mephestys:
The navbar clearly wants to hurt me, as I can't seem to work with it when I try any width settings that aren't hard pixels/em's/etc. (meaning that, when I use percentages, the links just bunch up with one another and stop responding to further attempts to justify or align them)

mixelpixel:
@Mephestys ya, those navbars are a a **surly** bunch! 😉 

Josh is going over percentages tomorrow and a vidy is posted on Piazza and in Slack forya!

PR #4, opened 2018-01-31
Aaron Burk

Arkoma:


tsj7:
Hey Aaron, project looks good. Nesting, and organization is spot on. I like that you committed when you were done with mobile and then again with tablet. That organization structure is a good one to follow.

For ease and consistency if you want to put the title of the pull request with just your name that would help us organize them and also takes the work of having to come up with a name, plus the name of the pull request may be true for the current code block you are working on but later in the day or the next day it may be different. Definitely still use context specific naming for your commits though. Good work man.

PR #7, opened 2018-01-31
 Shobana Ramesh @phone

shobanavv:


wlharvey4:
Shobana:
Nice job; would like to see your commits increase in number and the commit messages given more detail.  
1) Each git commit message should be structured so that the first line contains a title or description of the problem being solve or the issue being worked on;
2) 2nd line left blank
3) 3rd and subsequent lines should be detailed descriptions of the files modified, a summary of how they were modified and why, and the problem solved.

Getting into a habit now with your commit messages will make them a lot better by the time they matter, and when you go into a job and have great commit messages, you will be a valuable member of a team.  The people reviewing your commits and pull requests don't have a lot of time, and if you can give them detailed information quickly, they will know if the pull request is something they want, and where there might be problems, and they can focus their attention quickly.  

Wesley

PR #9, opened 2018-01-31
Dani Tacheny: finished with percentages

danitacheny:


wlharvey4:
Dani,
Awesome work; lots of commits, with some informative titles (for the most part).

Something to make your commits and pull requests a little better:
1) Git is set up so that the first line of your git message is a title or description of the problem being addressed
2) Second line should be left blank
3) Third and subsequent lines are for detailed descriptions of the files affected and a summary of the changes made.
4) The more informative you can make your commit messages, the faster reviewing members of your team can determine what you have done, why, and whether it is something they want, and will point out potential problems.  They do not want to have to go through your code line by line unless they are looking  for something, and your commit messages help them figure out where to focus their energy.
5) So, for example, the message "phone looks okay" could be titled "Fix phone issue of <description".  [2nd line blank] "file 1:  made a change in line .. to do this; file 2: adjusted margin left; file 3: ..."  And so forth.

You are really close to great commit messages (the best I've seen so far), so take it to the next level.

Wesley

PR #10, opened 2018-01-31
CS7 Leon <w-i-p>

IchabodPain:


wlharvey4:
Leon,
I've reviewed your commit.  In addition to the subject matter of the lesson, we are also teaching how to make meaningful Git commits and pull requests.  Obviously, this lesson is very subjective, and your work looks good.

On the subject of commits: you should be committing often.  Git is set up so that the first line of a commit message is seen in a history log, and so should be titled with a description of the main problem addressed by the commit.  The second line should be left blank.  The third and subsequent lines should have detailed listings of files and changes made to them and the purpose of the changes or the problem being addressed.  This method of making commit messages needs to be developed into a habit.  I know right now it must seem a pain, but this type of work is necessary for those reviewing your pull request to quickly understand the changes you are introducing and if it is something they want or if it might cause problems.  It is a good idea to at least make each commit message conform to that form, and when it becomes important, it will be a fine-tuned habit.

Wesley

PR #13, opened 2018-01-31
Clara Andrew: Responsive-web-design

clarakosi:


wlharvey4:
Clara,
Nice job; I've looked through your commits; everything looks good.  Thanks for putting your name in the PR request.

Some thoughts for making commits:
1) Git is set up so that the first line of your commit shows up in a log statement, and is designed for a title or topic or problem being addressed.
2) Second line should be left empty
3) Third and subsequent lines should be used for detailed explanations and problems being address.  It's a good idea to detail the files changed, and describe how they were changed and why.

Obviously, you have more on your plate in these sprints than taking a lot of time for detailed commit messages, but its a good idea to have in mind what is expected, and then try to make it a habit.  Good commit messages are vital when you are working in production, as it helps the reviewers really understand what changes you are introducing and if that is something they want or expect.  

Wesley

PR #26, opened 2018-02-01
Michael Marshal

mikemarshal:
It's kind of working but when the screen is sized down the container seems to stick to the left instead of centering itself (even with margin: auto)

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -165,8 +165,9 @@ h5 {
 }
 @media (max-width: 768px) {
   .container {
+    position: absolute;
     margin: 0;
-    width: 100%;
+    width: auto;
   }
 }
 header {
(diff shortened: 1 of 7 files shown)


mikemarshal:
fixed it by adding position absolute for the smaller than laptop widths (elements unapply when it's not that media specifications?) ... not sure why I need width auto vs width 100% to fix this... the size of the body container seems to get disproportionate and smaller as the size decreases while the container scales as it should (why?)

tsj7:
We chatted in slack about this but figured I'd put it here to. Maybe not an ideal solution but seems so help functionality:

```body {
  height: 100%;
  font-family: 'Titillium Web', sans-serif;
}```


```.container {
  width: 100%;
  max-width: 880px;
  margin: auto;
}```

PR #31, opened 2018-02-01
Ben Campbell CS6 - Responsive Web Design

brcampbell001:
…sick


manishavi:
Hi Ben,
Thanks for putting your name on the pull request.The mobile view looks good.
Suggestion:
It's a good practice to do git commits often and with useful comments.
Thanks.

PR #33, opened 2018-02-01
Dixie Korley - Percentage Complete

DixieKorley:


antoniomelendez:
Hi Dixie, 
Great work, it looks like everything is responsive as it needs to be. Fabulous job on committing as you made progress in the assignment. My only suggestion would be, add your name to the pull request.

dk574:
Thank you

PR #39, opened 2018-02-02
Igor Yermak

didacta:
messed up somewhere when changing the <p> for phone accessibility. Getting the concepts down well but I think I might need to start over, my folder structure is all messed up due to the LESS watcher making a duplicate less folder in the css tree

### DIFF
--- a/css/less/contact.css
+++ b/css/less/contact.css
@@ -1,0 +1,4 @@
+/* Contact Styles */
+.contact p {
+  margin: 40px 0;
+}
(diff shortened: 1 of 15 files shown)


tsj7:
Hey Igor,

If you are getting the concepts down that is what is important. I see what you are saying about some of the files being duplicated. I'm not sure how that happened. In the future one thing that could help is commiting more often, that way you can reference your files at different stages and compare changes. Putting commits with descriptive titles can also help. Overall nice work!

Tyge

PR #44, opened 2018-02-02
Reponsive Design Day II Submit

csparks878:
Everything works except the <li>s at the top. Asked for help but it seems the TAs are out for the night.


tsj7:
Hey Charlie,

Your code looks great. Well organized and laid out. Only suggestion would be to commit a bit more often. Each time you finish a module or hit a checkpoint it isn't a bad idea to commit, you can then revert to a previous commit if needed. Also in your comments if you could provide a bit more detail and context it could help too. Not sure what was meant by "everything working except the s at the top".  But overall really good job Charlie!

PR #51, opened 2018-02-07
Brett Sullivan  - Fully Responsive

swedishgoodbye:


css/index.css
```diff
@@ -174,6 +174,57 @@ footer {
     margin: 40px 0 20px;
 }
 
+@media(max-width: 768px){
```
StevieIsmagic: Comments are your friend! It may help others / you to comment the kind of device you are targeting with this media query.

css/index.css
```diff
@@ -174,6 +174,57 @@ footer {
...
+
+    .contact{
+        display: flex;
+        flex-flow: column;
+        align-items: center;
+    }
+
+}
 
 @media(max-width: 412px){
```
StevieIsmagic: Same here.

PR #56, opened 2018-02-07
Challenge 2 Complete - Daniel Plott - Responsive Web Design

DeepLot:
Completed Challenge 1 - Responsive Web Design - No horizontal scrollbars at break points and adjusted font sizes and flow for mobile. Stretch goal completed in new-index.css and new-index.html

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -64,13 +64,15 @@ h1, h2, h3, h4, h5 {
 }
 
 .container {
-    width: 880px;
-    margin: auto;
+    max-width: 880px;
+    width: 100%;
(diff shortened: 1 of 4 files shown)


DeepLot:
Adjusted all hard px values to %'s and increased responsive access for application. Made minor adjustments to top content flow and some margin/padding issues of class cta-text. Will continue to improve.

jryanp1:
OK looks good. Some minor things in console but those only showed up when I used github htmlpreview, it just doesn't like or couldn't find something so it squawked about that. But the breakpoints look right.

PR #64, opened 2018-02-07
Drew Moody Completed Responsive Web Design - Project One

DrewMoody:
Updated the index.css file with changes to make the page more responsive at the 400px, 768px, and 1100px levels. Going to work on a mobile-first design next.

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -60,7 +60,7 @@ html, body {
 h1, h2, h3, h4, h5 {
     font-family: 'Bangers', cursive;
     letter-spacing: 1px;
-    margin-bottom: 10px;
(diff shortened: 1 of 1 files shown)


DrewMoody:
Ended up adding in:
`@media (max-width: 768px)
index.css:249
.cta img {
    margin-bottom: 1%;
}`
after pushing just to give the img a bit of space once it stacks on top of the h1. Not going to bother pushing up again for that one change.

DrewMoody:
Noticed a few small margin fixes were needed so made an adjustment there. Also forgot to convert the navbar from px to rem so converted that.

PR #67, opened 2018-02-07
Jeremy Kepf  PROJECT 1 max-width COMPLETE

Khepf:
I'll work on the min-width stretch goal if there is time but I need a break.

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -55,6 +55,7 @@ table {
 html, body {
     height: 100%;
     font-family: 'Titillium Web', sans-serif;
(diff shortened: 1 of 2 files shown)


Khepf:
Thank you Randall.

In regard to the pull requests, I'll start each project with a pull request
and add commits to it from this point on. Other TAs were saying to wait
until the final submission to pull, which is why I was waiting.

Thanks again! First week through and I couldn't be happier with the level
of challenge and the results.

Regards,

Jeremy

PR #72, opened 2018-02-08
Project1/Project2--WIP Sarika Mirchandani

sarikaricky:
hello. min and max queries as per related to .container.


darkness198:
Sarika, pulled down your project and attempted to resize and saw there was no horizontal scroll bar down to the phone resolution. However, just sizing with percentages makes it lose the designed look we want. Technically it works, but try to play around with media queries to get more experience with side projects maybe to layout your own design. No console errors, clean code, don't forget to put your name on the PR!

PR #74, opened 2018-02-08
Boniface just did tablet view im working on phone right now

bmakau:


css/index.css
```diff
@@ -164,12 +169,67 @@ header .logo {
...
+}
+
+/*Tablet*/
+
+@media(max-width:768px){
+    .container{
+        width:100%;
+    }
+header .logo{
+
```
randallwhitlock: if you implement a percentage-based width on the logo image, it will smoothly resize as the viewport size changes, which is great for transitioning between one break point and the next.

PR #76, opened 2018-02-08
Megan Williamson | Responsive Web Design Project 1 | Completed

gooseandmegander:


jryanp1:
Ok good progress. There is a horizontal scrollbar at about 900px so look at that.

css/index.css
```diff
@@ -107,30 +114,46 @@ header .logo {
...
     display: flex;
     justify-content: center;
     flex-flow: column;
     align-items: center;
-    margin: 0 82px 0 0;
+    margin: 0 9% 0 0; /* 82px/880px = 9.32% */
+}
+
+.cta .cta-text button {
+    border: 1px double black;
```
sarafritz88: Even though we leave the button width alone, we can change other things using px to be using rems or percentage.

PR #79, opened 2018-02-08
Kaitlyn Schorr - Complete - Used Max-width and media queries

kait-schorr:
There are multiple commits because I found mistakes after I submitted. Apologies!


jryanp1:
OK, good progress. No major console errors (other than in github preview which gives everybody errors). A horizontal scrollbar does appear below about 1100px page width so take a look at that.

PR #80, opened 2018-02-08
added Jake Henry assignment day3

Blast3d:


mvpache:
Great job using all the breakpoints plus REM and %.

PR #84, opened 2018-02-08
WIP - Shaun Kelly - Responsive WD - Project1

RedSkelly:


RedSkelly:
Completed Project1. Renaming pull request to "Project2" to work on same file.

Questions/Concerns:
- how can I include the img w/o using `display: none;` or using `flex-direction: column;` ?  (see comment in index.css).

- The index.html, when opened in a browser, views perfectly at both 400px and 768px. However, when w/in this range, the page looks janky, it doesn't format correctly.

PR #88, opened 2018-02-08
Sumayyah's Responsive-Web-Design

sumi419:
Changed the header for mobile, tablet, and laptop view using flex wrap.

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -3,65 +3,7 @@
    License: none (public domain)
 */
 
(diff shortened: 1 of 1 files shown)


sumi419:
WIP

47analogy:
Sumayyah, your page is responsive at the breakpoints, you've converted to rem, and incorporated % into the layout. Are there any particular concepts that you are unsure about? If so, please review the lecture/solution videos and reach out with your concerns. Great job!

PR #90, opened 2018-02-08
William Pelton Complete

runranron:


jryanp1:
OK good progress. We have a horizontal scroll bar at under 900 px width, so look at that. No console errors other than it doesn't like using of document.write but github squawks error at everybody so confirm with your teacher on that.

PR #92, opened 2018-02-08
Responsive-Web-Design Project: Jeremy Jones

crypto-jones:


darkness198:
It looks like you understood this one well, no horizontal scrolls bars when I pull down your project and open it in chrome. No console errors and clean code. Only thing is that Great Idea img is hanging out there on the right side in between 400 and 768 px, other than that looks good. Don't forget to put your name in the PR, thanks!

PR #99, opened 2018-02-08
Redesign Michal Chmielniak

mgchmiel:
I think I did this correctly my computer did not respond to well with the changing of pixel sizes.

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -139,7 +139,7 @@ table {
 			width: 50%;
 		}
 	}
-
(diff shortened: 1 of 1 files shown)


47analogy:
Michal, This appears to be a WIP as far as the responsiveness at 400px and768px,. Did you get a chance to watch the lecture/solutions video yet? If you did and there's something that's still not clear or confusing, please let us know and we'll get you the resources that you need. Good job!

PR #102, opened 2018-02-08
Josh Coyne wip responsive-web-design project

joshuacoyne:


47analogy:
Josh, Your pages is responsive at the breakpoints and you've incorporated a percent-based layout. Your commit messages are detailed (appreciate it!). I would watch the solutions lecture video to gain some more insight on this project and then ask about anything that still gives you trouble. Great job!

PR #109, opened 2018-02-09
nour hassan

nourmhassan:


mvpache:
Still a scroll bar at phone size but looks great outside of that. Play with element sizing.

PR #124, opened 2018-03-07
Jared Cuffe

jcuffe:
Drastically changing the layout using media queries was very straightforward and I was able to finish this project very quickly. I am looking forward to using percent(%)-based layouts, as it felt a bit awkward to be zeroing out pixel margins inside of the media query sections!

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -64,8 +64,7 @@ h1, h2, h3, h4, h5 {
 }
 
 .container {
-    width: 880px;
-    margin: auto;
+    width: 100%;
 }
 
 /* Header Styles */
@@ -73,18 +72,30 @@ h1, h2, h3, h4, h5 {
 header {
     width: 100%;
     display: flex;
+    flex-direction: column;
+    align-items: center;
 }
 
 header nav {
-    margin-top: 40px; 
-    width: 610px;
+    width: 100%;
     justify-content: space-between;
     display: flex;
+    flex-direction: column;
+    align-items: center;
 }
 
 header nav a {
     text-decoration: none;
     color: #b0b0b0;
+    font-size: 110%;
+    width: 100%;
+    padding: 20px 0px;
+    text-align: center;
+    border-top: 1px solid black;
+}
+
+header nav a:last-child {
+    border-bottom: 1px solid black;
 }
 
 header nav a:hover {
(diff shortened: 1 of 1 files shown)


kiadorii:
Awesome commit messages and thank you for the comment. Small details like these make a world of a difference!

css/index.css
```diff
@@ -136,40 +150,105 @@ header .logo {
...
 }
 
 /* Footer Styles */
 
 footer {
     text-align: center;
     margin: 40px 0 20px;
+}
+
+@media (min-width: 768px) {
```
kiadorii: Glad to see you implementing min-width and percentages. You are already ahead of the curve and hope you continue to do so!

kiadorii:
Overall your code looks great. There are only a few acute details that I would want you to look at using the PerfectPixel extension on google chrome. The spacing is good and in regards to your concern about height in percentages and rem, you are already using rem for font-size which is good. You would only use percentages for margins and spacing. Like you mentioned before in the chat, the percentages need to be based off of something. 

Here is a few sources that I would like for you to gloss over if you wanted to understand it a little better:
https://stackoverflow.com/questions/22288494/margin-top-in-percentage-not-working-as-expected
https://www.impressivewebs.com/vertical-percentages-css/

I hope this was helpful!

css/index.css
```diff
@@ -93,36 +97,36 @@ header nav a:hover {
...
-    letter-spacing: 5px;
     display: flex;
     justify-content: center;
     flex-flow: column;
     align-items: center;
-    margin: 0 82px 0 0;
+}
+
+.cta .cta-text h1 {
+    font-size: 7.5vw;
```
kiadorii: Impressive to see you using vw, continue going further than the knowledge that is given to you in lecture! 👍

css/index.css
```diff
@@ -131,45 +135,133 @@ header .logo {
...
 
 footer {
     text-align: center;
-    margin: 40px 0 20px;
-}
\ No newline at end of file
+    margin-top: 40px;
+}
+
+@media (max-width: 600px) {
```
kiadorii: Nice work with the break-points. You had brought up a key point where there were some weird funk going on as you resized the browser. It definitely looks a lot more fluid.

css/index.css
```diff
@@ -93,36 +97,36 @@ header nav a:hover {
...
-    margin: 0 82px 0 0;
+}
+
+.cta .cta-text h1 {
+    font-size: 7.5vw;
+    letter-spacing: 5px;
+    font-weight: bolder;
 }
 
 .cta .cta-text button {
```
kiadorii: One thing to note for this button is that as I got to the phone size 400, the getting started and border are cutting off on each other.

PR #126, opened 2018-03-07
Starting on tablet size changes

CruiseBrantley:


kiadorii:
Thank you for the clear commit messages! Love seeing the progression.

css/index.css
```diff
@@ -206,35 +206,73 @@ footer {
...
+        padding:10px;
+    }
+
     .main-content img {
         width: 100%
     }
-}
-*/
\ No newline at end of file
+}
```
kiadorii: I hope the solution lecture was helpful and that it was able to get you to understand more than you already know!

css/index.css
```diff
@@ -174,28 +174,50 @@ footer {
...
+    .main-content img {
+        width: 100%
+    }
+
+    .contact {
+        min-height: 180px;
+    }
+
+    footer {
+        align-content: flex-end;
```
kiadorii: I think the error you were getting from your footer may be how it was already implemented beforehand. Check out the Dev Tools on Google Chrome. When I clicked on the footer, it was only receiving it's css from .p.

css/index.css
```diff
@@ -57,15 +57,21 @@ html, body {
...
 
+html {
+    font-size: 62.5%;
+}
+
 h1, h2, h3, h4, h5 {
     font-family: 'Bangers', cursive;
     letter-spacing: 1px;
-    margin-bottom: 10px;
+    margin-bottom: .3%;
```
kiadorii: As mentioned in yesterdays standup/solution meeting, top and bottom margins should be in px vs %. They are more for aesthetic properties rather than responsiveness 😄 But on saying that, the fact that you were able to get it to be responsive is awesome!

css/index.css
```diff
@@ -164,12 +175,177 @@ header .logo {
...
+    }
+    .contact{
+        font-size: 1.6rem;
+        padding-left: 30px;
+    }
+    .text-content {
+        padding:10px;
+    }
+
+    .main-content .middle-img{
```
kiadorii: For some reason, even though you have done an align, this image is set on the left. See what happens when you implement a margin-left here. And this goes for the horizontal lines as well 👍 Overall, really solid job on this project! I feel as if your confidence has grown so much from just this week alone.

PR #127, opened 2018-03-07
Natalie Schennum

Nschennum:
Just a pull request per request.


kiadorii:
I was SO happy to hear about how much you learned already and on top of that, it has only been a couple of days! Something that I would recommend is that, instead of writing your commit message as 'update' you may want to clarify what changes you have made exactly. For example, 'git commit -m "Progress on table for Project 1"`. This is taking from the feedback that your peer, Lukasz, mentioned of having clearer code. It should be that when you look back at your commit messages, you will not have a difficult time remembering what you exactly did.

css/index.css
```diff
@@ -172,4 +172,93 @@ header .logo {
...
+    }
+    .cta .cta-text{
+        margin: 15px; 
+    }
+    .main-content .middle-img {
+    	display: flex; 
+    	width: 100%;   
+    }
+}
+@media (max-width: 450px){
```
kiadorii: I believe the max-width should have been 400px.

css/index.css
```diff
@@ -172,4 +172,10 @@ header .logo {
 footer {
     text-align: center;
     margin: 40px 0 20px;
+}
+meta (max-width: 768px){
+    .container {
+        width: 100%; 
+        padding: 25px;
```
kiadorii: This is the right path to your project. I saw that in your latest commits that you were able to implement the solution code. I hope that it was helpful to you in being able to understand the material more in-depth. Keep up your awesome work and positive attitude! ❤️

css/index.css
```diff
@@ -51,7 +46,9 @@ table {
 * {
     box-sizing: border-box;
 }
-
+html { 
+    font-size: 62.5%;
```
kiadorii: Important to note that this is what allows font sizes to be changed to rem in base of 10.

kiadorii:
I hope the solution code has been helping you gain insight to this project. The general rule of thumb is that if you ever get stuck on something for 20 min, ask the channel! You seem to be getting a better grip of the course and I hope you believe that too!

Nschennum:
did I not set the font size in html properly?

PR #128, opened 2018-03-07
 768 screen finished working on 400px now

kyle-kauzlarich:


DonnieU:
All 3 views are very close to design. Great use of rem for font size and percentages for widths. Nice, clean code. Keep up the great work!

PR #130, opened 2018-03-07
Axhon Ruiz RWD

axhon:


kiadorii:
Thank you for the clear commit messages!

css/index.css
```diff
@@ -328,6 +325,26 @@ header .logo {
   padding-right: 20px;
 }
 
+@media (min-width: 769px) {
```
kiadorii: Impressed with your implementation with min-width! You're definitely ahead of the curve here and hope you continue to do so!

css/index.css
```diff
@@ -177,37 +192,38 @@ header nav a:hover {
 }
 
 /* Desktop Header */
-@media (min-width: 769px) {
-  .container {
-    width: 880px;
-  }
 
+@media (min-width: 769px) {
```
kiadorii: Love seeing the implementation for min-width for this project. They definitely come in handy and stops the cascading from spilling over to a section you would not want.

css/index.css
```diff
@@ -177,37 +192,38 @@ header nav a:hover {
...
   }
 }
 
 /* Tablet Header  */
+
 @media (max-width: 768px) {
   .container {
-    width: 100%;
-    margin: 0 20px;
+    max-width: calc(768px - 6%);
```
kiadorii: Neat stuff implementing calc, keep at it!

css/index.css
```diff
@@ -218,50 +234,49 @@ header nav a:hover {
...
-    width: 400px;
+    width: 100%;
     text-align: center;
     border: solid #808080 1px;
   }
-
   header .logo {
     align-self: center;
-    margin: 25px 0px 0px 20px;
+    margin: 6.2% 0% 0% 5%;
```
kiadorii: I mentioned this in the solution lecture for Thursday night, but the general rule of thumb is that the top and bottom margins are to be in px while the sides are in percentages. Top/bottom are more of aesthetic properties rather than for them to be fluid/responsive. Upon seeing that, making everything into percentages can be quite tedious so kudos to that! You have a lot of great implementation from new material that haven't been covered to just having a great looking site.

PR #132, opened 2018-03-07
Completed Responsive Web Design

ShawnStewart:


DonnieU:
All 3 views look great and are very close to design spec. Nice inclusion of extra media query at 725px. Proper use of rem for font sizes and percentages for margins. Good, clean code. Keep up the great work!

PR #133, opened 2018-03-07
Harrison Brock (RWD)

harrisonbrock:
I have 768 done and I'm working on 400px

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -202,8 +202,11 @@ footer {
 @media (max-width: 400px){
 
     hearder {
+        display: flex;
         flex-direction: column;
-        align-items: center;
+        justify-content: center;
+        width: 100%;
+   
     }
 
     header nav {
@@ -220,9 +223,10 @@ footer {
     }
 
     header .logo {
+        flex-direction: column;
+        justify-content: center;
+        margin: 0 0 0 0;
         order: -1;
-        margin: 20px 0;
-        align-self: center;
     }
 
     .cta {
--- a/css/index.css
+++ b/css/index.css
@@ -55,6 +55,7 @@ table {
 html, body {
     height: 100%;
     font-family: 'Titillium Web', sans-serif;
+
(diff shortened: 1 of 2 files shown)


kiadorii:
Love your commit messages, continue what you're doing for the future PR's!

css/index.css
```diff
@@ -76,8 +81,10 @@ header {
 }
 
 header nav {
-    margin-top: 40px; 
-    width: 610px;
+    /*margin-top: 40px; */
+    margin-top: 4.5%;
```
kiadorii: As you can probably recall from the meeting yesterday, margin top/bottom would almost always be in px since it is more for aesthetics while left/right will be in percentages.

css/index.css
```diff
@@ -158,18 +175,137 @@ header .logo {
...
+    /*padding-right: 20px;*/
+    padding-right: 2rem;
 }
 
 /* Contact Styles */
 
 .contact p {
-    margin: 15px 0;
+    /*margin: 15px 0;*/
+    margin: 1.5rem 0;
```
kiadorii: **This applies to several areas in your code** If you can, refrain from using rem for margins. Not necessarily bad but rem is dependent on the font-size while margins are dependent on the parent it is called on.

css/index.css
```diff
@@ -158,18 +175,137 @@ header .logo {
...
+        flex-wrap: wrap;
+        height: 75%;
+    }
+
+    .main-content .bottom-content {
+        flex-wrap: wrap;
+        align-content: space-between;
+        height: 110%;
+    }
+
```
kiadorii: Overall, really solid job! Even getting the top/bottom margins to look good using percentages is a struggle and in you being able to do just that is great to see. Also, love seeing the fact that you commented your margins that were in px rather than deleting them whole. It is a good way to compare any changes that you would like to make! 👍

PR #135, opened 2018-03-07
Responsive Web Design Assignment 1

cjmarsh725:
Created adaptive styling for the website using media queries and reference design documents.

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -234,5 +234,17 @@ footer {
 		margin: 0 auto;
   }
 
+	.cta {
+    flex-direction: column;
+	}
 
+	.cta .cta-text {
+    align-self: center;
+    margin: 0 0;
+	}
+
+	.cta img {
+		margin-top: 20px;
+		align-self: center;
+	}
 }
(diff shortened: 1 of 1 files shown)


css/index.css
```diff
@@ -172,4 +172,91 @@ header .logo {
...
+
+	header nav a {
+		height: 30px;
+		border-style: solid;
+		border-top-width: 1px;
+		border-color: gray;
+		align-self: stretch;
+		text-align: center;
+	}
+
```
frogr: Weird inconsistencies with the white space above and below this point. You use Atom right? I use Prettier with ESLint and it helps a ton with stuff like this.

css/index.css
```diff
@@ -172,4 +172,91 @@ header .logo {
...
-}
\ No newline at end of file
+}
+
+/* Tablet formatting */
+
+@media (max-width: 920px) {
+  .container {
+    max-width: 880px;
+		width: calc(100% - 40px);
```
frogr: I think `width: 100%;` works fine, but this is a neat approach.

PR #141, opened 2018-03-07
completed challenge

phantomflynn:


sulemc:
End products look great. The tablet's nav bar needs a left margin. As for your code, remember media query css goes at the bottom so that it inherits the properties above it. Although having it throughout the CSS works, it's not best practice. You should have one media query per px constraint and all the changed CSS inside that query. If you have any questions about that - let me know!

PR #147, opened 2018-03-07
responsive

faulkenfett11:


DonnieU:
All 3 views are near pixel perfect and definitely meets design spec. Great use of rem for font sizes and percentages for ALL margins. Good, clean code. Fabulous job!

PR #148, opened 2018-03-07
Keith Haag - Incomplete RWD

kkhaag:


css/index.css
```diff
@@ -172,4 +172,71 @@ header .logo {
...
+
+img {
+    max-width: 100%;
+    height: auto;
+}
+embed,iframe,object,video {
+    max-width: 100%;
+}
+
+
```
HTMLGhozt: I believe you're missing a closing bracket `}` around here.

css/index.css
```diff
@@ -172,4 +172,71 @@ header .logo {
...
+.cta .cta-text {
+    margin: 0px;
+}
+
+img {
+    max-width: 100%;
+    height: auto;
+}
+embed,iframe,object,video {
```
HTMLGhozt: There shouldn't need to be a reason where you need to alter any of these particular elements.

css/index.css
```diff
@@ -172,4 +172,71 @@ header .logo {
...
+    justify-content: space-around;
+  }
+ .logo {
+    max-width: 100%;
+   border: 3px solid pink;
+    height: auto;
+    display: flex;
+   justify-content: center;
+   align-items: center;
+}
```
HTMLGhozt: on line 91 we're applying margins to to the logo:
```
header .logo {
    align-self: flex-end;
    margin: 28px 0 0 91px;
}
```

css/index.css
```diff
@@ -172,4 +172,71 @@ header .logo {
...
+    max-width: 100%;
+}
+
+
+
+@media (max-width: 400px){
+    body {
+    width: 100%;
+    height: auto;
+    padding:;
```
HTMLGhozt: empty value on padding attribute.

PR #151, opened 2018-03-07
Nalee Riddell

odax:
Partially completed J-Query for 400px break. Properly set up header


justinborek:
Nice work on this one, you got super close. Probably the hardest part of this one is figuring out the navbar switching to vertical on the mobile view, and you got that. Other than tweaking margins all around to add padding, you got it. One thing to work on though is making sure your code is cleaner (i.e. watch for incorrect tabs/spaces and newlines). Good commit messages. Keep it up!

PR #156, opened 2018-03-08
jacob Lyerla

JacobLeonLyerla:


DonnieU:
All 3 views are very close to design except footer text size is too small for 400px. Otherwise everything looks great. Appropriate use of rem font sizes and percentages for widths. Good looking code and a great job!

PR #157, opened 2018-03-08
Progress on RWD Project 1 ran out of time to fix .cta img and main…

ktan114:
…-content text boxes.


frogr:
Hey Kevin! This solution isn't quite satisfactory. There's still horizontal bars, and so it's not technically 100% responsive. You're correctly breaking down as the view gets smaller, but this happens: 
image

when it needs to be more like this:
image

As a quick check off the top of my head, I think this has to do with you hardcoding the width pixels instead of having them take like 100% or another %-based property. If you have time, definitely refer to the solution code and make sure you take note of the big changes within the media queries.

PR #159, opened 2018-03-08
Responsive-Design - Justin Mathew

jekm321:


justinborek:
Nice work with this one! Just a couple tweaks to margins on the tablet view and it's done! As always, super clean code, good use of comments. Keep up the great commit messages as well. Nicely done!

PR #161, opened 2018-03-08
Responsive Design Challenge I complete

abrambueno1992:
Responsive Design Challenge I complete


frogr:
Hey Abraham! Looks pretty good, but you still have some horizontal scrolling going on at each break point. This has to do with the content of the page extending past the 768px/400px boundary, and tends to occur when you hardcode pixels instead of using %s. 

Your breakpoints look solid, and it's all the styling maintains without breaking. I also loved that you added in an extra breakpoint! Not necessary, but very cool. Nice work

PR #162, opened 2018-03-08
Imran_Ahmed_ResponsiveWebDesign2

imsdhk:
with 768 and 400

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -44,233 +44,217 @@ h1, h2, h3, h4, h5 {
 
 /*---------------desktop view---------------*/
 
-	body { 
-		font-size:62.5%; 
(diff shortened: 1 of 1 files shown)


frogr:
Hey Imran! It looks like you didn't quite get rid of the horizontal scrolling, so I wouldn't really be able to classify this as entirely responsive yet. Check the width of your containers! You should not be hardcoding pixels there. Use a %. Definitely refer back to the solution code and compare!

There's also a few aesthetic flaws, the button breaks (the text overflows out of it) and the nav borders don't quite look like they do in the spec. If you have time today, consider going back and checking these things!

Let me know if there's anything I can do to help!

PR #169, opened 2018-03-08
Great project

ldavis0866:


justinborek:
Nice work with the one! I can tell things are catching on as we go. About the only things missing from your original solution were whitespace around the layout (which per our lecture you can control with box model things like `margin` and `padding`. I saw you also originally had the image in the navbar on the bottom. I think the `order: -1` we demonstrated in lecture should also help there. Great progress, keep up the good work!

PR #171, opened 2018-03-08
Subra Yegna

Sy-CoDe:
latest changes to responsive design assignment. Still needs fixing. lot more practice needed to understand & execute


justinborek:
Nice start on this one. Your tablet implementation looked good. I saw you had commented out the code from your mobile breakpoint, was that just the solution code to play with? Either way I think you were on the right track, and would have had it given more time. Keep it going! Also remember to submit those PRs! :smile:

PR #173, opened 2018-03-08
Hilal - RDW

HTMLGhozt:


css/index.css
```diff
@@ -175,19 +175,109 @@ footer {
...
+/* media queries */
+
+/* tablet*/
+
+
+@media(max-width: 768px){
+    .container{
+        width: 100%;
+        margin: 10px 200px 100px 10px;
+
```
HTMLGhozt: It would have been nice to see this as padding instead of margin. Margin slightly pushes elements off the side when looking at 768px view.

css/index.css
```diff
@@ -175,19 +175,109 @@ footer {
...
+    }
+
+    .cta-text{
+        padding-left:50px;
+    }   
+
+    footer{
+        margin-top:70px
+    }
+
```
HTMLGhozt: Where's the adjustments for `top-content` and `bottom-content`?
top-content bottom-content flexing

css/index.css
```diff
@@ -175,19 +175,109 @@ footer {
...
+      
+       
+   
+   }
+
+   .logo{
+       order:-1;
+       padding:5px;
+       border-bottom: 2px solid black;
+       width:100%;
```
HTMLGhozt: By setting the width to 100% we force the width to be the full 400px, which bloats the logo.
Logo

css/index.css
```diff
@@ -175,19 +175,109 @@ footer {
...
+  }
+
+  .cta{
+      display: flex;
+     justify-content: space-around;
+     width:99%;
+     height: 385px;
+
+  }
+  .cta-tex{
```
HTMLGhozt: `.cta-text` is misspelled.

css/index.css
```diff
@@ -175,19 +175,109 @@ footer {
...
+        display:flex;
+        flex-direction: column;
+        flex-wrap:wrap;
+        height:800px;
+        padding-right:5px
+     
+    }
+
+    .cta-text{
+        padding-left:50px;
```
HTMLGhozt: Why are we adding padding instead of centering with flex?

PR #179, opened 2018-03-08
my responsive web design

macknwandu:


sulemc:
Great work, especially considering you had a migraine!
The desktop/tablet have some nav bar issues. And the tablet .cta elements aren't correctly aligned.
If you have time this weekend see what you can do with those.

PR #183, opened 2018-03-09
Jonathan Greene -- Got most of it done I feel like just missing some stuff both in tablet and android. 

Ashiko124:


Ashiko124:
fgdf

PR #189, opened 2018-04-04
Cassandra Lamendola

CassLamendola:
Project 1: Adaptive Layout

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -174,6 +174,10 @@ footer {
     margin: 40px 0 20px;
 }
 
+/************************
+**Breakpoint for tablet**
+************************/
+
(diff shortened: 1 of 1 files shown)


dresean:
The web page looks really awesome! Some of the media queries for mobile have a different scale/design compared to other devices. For example if you use the chrome dev tools and switch devices in their interface, not all the mobile resolutions have the same UI, while that is not a problem as far as design, just be mindful of gauging your measurements to be consistent across smaller and larger mobile devices. A google pixel should have the same UI as an iPhone 6s. The coding style is very consistent throughout. Also the commenting and commits are looking really good. Keep up the good work!

PR #191, opened 2018-04-04
Alex Dykas

udykas:
CS 10 Responsive Design Project 1


etisdew:
Oh man you got sooooo  close, @400 please get the .cta down to size and PM me to let me know you're done. Sooooo close, I rather see you finish it. This would be a great item for the end of day through. :+1:

PR #204, opened 2018-04-04
Dil Sombayeva RWD

Dilstom:


dresean:
This project is a solid piece of code and execution. 100% responsive! Some issues in the commit messages but that is nothing that can’t be fixed with time. Be sure to include more specific messages when adding commits. Also when branching please make sure to include the name of the current assignment as the branch name. The media queries are looking good! Some small amounts of work can be made to smooth the transitions from bigger to smaller and vice versa. The code style is also looking much cleaner this go round! The way you are prioritizing is great as well. Comments aren’t always needed but they are helpful in deciphering which section of the code we are looking at. Overall this is excellent work. Great job Dil!

PR #212, opened 2018-04-04
Responsive Design

mr-burton:
Modifying with media queries still needs work.

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -205,13 +205,26 @@ footer {
         width: 100;
     }
     nav {
+        max-width: 100%;
         display: flex;
         flex-direction: row;
+        order: 1;
         margin: 20px;
     }
     nav a {
+        display: flex;
+        flex-direction: column;
+        flex-wrap: wrap;
         padding: 20px;
-        order: 1;
+        order: 2;
+    }
+    .main-content {
+        max-width: 768px;
+        display: flex;
+        flex-direction: coulumn;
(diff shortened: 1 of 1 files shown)


designerexpert:
## Part 1: Adaptive Layout
* [ ] Breakpoint: 400px.
* [x ] Breakpoint: 768px.
* [ ] Review each design file and adjust your styles at each breakpoint.
* [x ] No horizontal scroll bars at the exact breakpoint.
* [ ] Flexbox layout techniques used for this approach

## Part 2: Fully Responsive Design
* [ ] Convert all fixed widths throughout the site into percents
* [ ] Convert all left and right margins and paddings into percents
* [ ] Convert all font-size pixels into rem units
* [ ] Make the breakpoints look pixel perfect using the design files
* [ ] Stretch Goal: Copy and paste your project to a new folder and make a 100% fluid layout design by removing breakpoints. There is no right answer here, I just want you to see what happens when constraints are removed. (Branches Can be used for this method)

## Project Manager's Thoughts
* Great job on the first break point!
* I would like to see some more work on the Tablet and Mobile view, specially the Background Color removed, and the bottom image targeted. If you get that done I will give you the extra points missing from part 1.
* Continue using meaningful commit messages.
* Push your code to GitHub as often as every hour, and remember to take a break after doing so for 5 minutes or so. Healthy body = Healthy Mind

PR #222, opened 2018-04-04
Project 3 by Diego 

dlinz7:
Project 3!

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -188,8 +188,7 @@ footer {
 
 
 
-        border: solid red;
-        background: gray;
+        
 
     }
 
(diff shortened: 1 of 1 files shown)


dresean:
The page is fully responsive! Amazing progress in such little time! Some of the media queries could’ve had a smoother transition when the device changes. Also there was an extra branch on the repository, I am not sure which one was the one you wanted to be reviewed. Upon viewing the branch labeled DL day 4, I noticed that the media queries were much better in their smoothness and the responsiveness of the design increased as well. Code style could use a little more work. Make sure to style and syntax check each code block before moving on to the next one. Overall this is yet another solid assignment and I look forward to seeing your code fu grow. Thanks Diego!

PR #227, opened 2018-04-05
Initial attempt to get the elements pixel-perfect. 

KayceHubbard:


JacquelynnRohrich:
image
So it's looking good, I could see a difference in this font-size and possibly the space around the lettering seem stretched, I would love for you to keep this good work up and perhaps this weekend get it pixel perfect! Especially since this is the week you will love with the styling.. Absorb it all :)

image

PR #228, opened 2018-04-05
not done yet

ahasch:


css/index.css
```diff
@@ -172,4 +172,38 @@ header .logo {
...
+    header{
+        width:100%;
+        display:flex;
+        flex-direction: column-reverse;
+        align-items: center; 
+    }
+    header nav{
+        width:100%;
+    }
+    header .logo{
```
etisdew: Almost, keep at it. Just mind the fact that its sliding with the scroll. I'm sure one of the class mates can help you get this.

css/index.css
```diff
@@ -172,4 +172,38 @@ header .logo {
...
+        display:none;
+    }
+    
+    .main-content .top-content {
+        width: 100%;
+    }
+    .main-content .middle-img{
+        width:100%
+    }    
+
```
etisdew: Let me know when you implement the @media max 400 so I can offer another review.

PR #241, opened 2018-04-05
Project 1 RWD Solutions

mag16:
changes maded to my solutions.


JacquelynnRohrich:
I just checked it out! Thanks again for promptly getting a pull request in. I was comparing to the solution code and it was so cool to see how warm or how cold you were to the exact pixel perfection.  I would recommend to keep yourself sharp to try to get it exact :) Hint.. You aren't that far off on most of it... Keep up the great work!

PR #273, opened 2018-04-06
Christina Marie Van Nostrand Kopecky

ckopecky:
Project 2  - Updated with Percentages and rems.

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -64,8 +64,8 @@ h1, h2, h3, h4, h5 {
 }
 
 .container {
-    width: 880px;
-    margin: auto;
+    width: 100%;
+    margin: 5%;
 }
 
 /* Header Styles */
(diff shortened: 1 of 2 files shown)


dresean:
Responsive Design 
The design looks awesome! In some of the break points the design looses its responsiveness, then when scaled to mobile size viewports it bounces back to responsiveness all except the image. Adjusting the main contents display properties will fix that. The  media query is nested nicely, with very little style error. In terms of reading your code, I would try to order the styles in the order in which they appear on the screen, that will help with readability and editing in the long run. Prioritizing properties like display, height, width within your code blocks will also help with debugging and readability as well. Commit messages looking good! Overall this is really solid work, and I am excited to watch your code fu get stronger each day! Thanks Christina!

PR #279, opened 2018-05-09
Emmeline: Responsive-Web-Design-Part I

emaquino44:


css/index.css
```diff
@@ -94,7 +98,7 @@ header nav a:hover {
 
 header .logo {
     align-self: flex-end;
-    margin: 28px 0 0 91px;
+    margin: 28px 0 0 11.3%;
```
JohnJustinn: Great use of percentages in the margin, and understanding how inheritance effects the particular percentage of each element being scaled.

css/index.css
```diff
@@ -106,19 +110,19 @@ header .logo {
...
     flex-flow: column;
     align-items: center;
-    margin: 0 82px 0 0;
+    margin: 0 9.31% 0 0;
 }
 
 .cta .cta-text button {
     border: 1px double black;
-    font-size: 16px;
+    font-size: 1.6rem;
```
JohnJustinn: Nice implementation of REM.  I can really tell you understand this concept and how it is applied.

css/index.css
```diff
@@ -172,4 +178,196 @@ header .logo {
...
+    width: 79.4%;
+    justify-content: space-between;
+    display: flex;
+    font-size: 1.4rem;
+
+}
+
+header .logo {
+    align-self: flex-end;
+    margin: 28px 6.51% 0 11.8%;
```
JohnJustinn: Perfect example of how the header should be scaled down.  Nice!

css/index.css
```diff
@@ -172,4 +178,196 @@ header .logo {
...
+}
+nav a {
+    padding: 15px 0 15px 0;
+    border: solid 1px lightgray ;
+}
+
+header .logo {
+    align-self: flex-end;
+    margin: 28px 0 0 20px;
+    order: -1;
```
JohnJustinn: Nice implementation of flexbox order here.

PR #282, opened 2018-05-09
LUIS MARTINEZ 

luiscmartinez:


bdurb:
@luiscmartinez Good job using commits to document your progress.  Code looks good, It looks like you missed the conversion to a percentage on the nav, but aside from that great work!

PR #283, opened 2018-05-09
still working on file

CookieMonsta89:


css/index.css
```diff
@@ -77,14 +80,16 @@ header {
 
 header nav {
     margin-top: 40px; 
-    width: 610px;
+    width: 69.318%; /*610px69.318*/
```
bdurb: I really like that you are using comments to show the changes you made from px to %.  This will be very beneficial for your learning progress.

bdurb:
@CookieMonsta89 Thank you for the PR.  Great use of commits, and specifically going into detail when you needed to on those commits.  Your code looks good, and % and rem are right on, looks great compared to the blueprint.  The only thing you could tweak are the header nav links.  After you switch to tablet view they are not responsive,  but thats a small detail.  Great work!

bdurb:
@CookieMonsta89 Great job on refactoring your code.  Looks great.  and good use of commits.  Keep up the good work!

PR #286, opened 2018-05-09
Blake Fletcher - Adaptive Layout

blkfltchr:


groov1234:
Looks good! Some minor changes for the phone and tablet @media queries. A few width's set to px instead of % based. I like how you changed the width of the header logo's container and centered the both.

PR #292, opened 2018-05-09
Erin Meredith - Responsive Web Design project 1

ekm79:


ekm79:
After adding the services design, I noticed my main tablet design changed. Working on fixing that...

ekm79:
still having trouble with header on the services tablet/mobile designs. Didn't alter header css at all for them.

PR #296, opened 2018-05-09
Made changes to RWD

monijoi:


bdurb:
Thank you for the PR.  Good job working through this challenge today.  There are elements that align very well.  There are some design's that are not exactly matching the blueprint, but there are also some that are quite close.  Keep at it!

monijoi:
Thank you!

PR #323, opened 2018-05-09
Libby Hart - cs11 [Responsive-Web-Design]

LibbyHart:


brcampbell001:
Hey Libby! I just reviewed your code. It looks pretty good when it breaks down to the different sizes. I see a few things that I feel that you could understand once the solution code is sent out in a few minutes. You're right, the code could be refined a bit and maybe a few unneeded items. But, the solution will show you how to do those refinements and Josh can reinforce that tomorrow.

PR #331, opened 2018-05-09
Adrian Adames CS11 - Responsive Web Design Project 1

adrianadames:


brcampbell001:
Hey Adrian! I reviewed your code and you are on the right track. The solution code should clear everything for you pretty easily. The margin you had set in container had everything shifting right. Remove that and it fits correctly. Other than that, it doesn't look bad at all. You are on the right track! Good job!

PR #336, opened 2018-05-09
Tom Camilleri Responsive Web Design I

tommycam:


groov1234:
Looks Great! I did notice one thing missing on the phone media query. margin was left undefined. If you define the margin in your phone ```@media``` you can get that pesky logo to center. Overall great work! Great job with specificity as well!

PR #359, opened 2018-06-14
Bryce Evans

BryceEvans:
My code, not Responsive


danitacheny:
Great work! The site looks great and your CSS is really well organized.

My only suggestion is to make more frequent commits with meaningful messages whenever you make a significant change to your code. The commit messages should make it clear what you've changed so you or anyone else who reads your code won't have to work very hard to understand what you've done.

Keep up the good work!

PR #371, opened 2018-06-14
Katherine Gilligan- Initial Commit

katiagilligan888:


danitacheny:
Everything looks great! The site looks great, your code looks clean, all good.

My only suggestion is the make more frequent commits with meaningful messages. I recommend making a commit every time you make a significant change in your code. This will make it way easier for yourself and others to understand what you've done!

Keep up the good work!

PR #373, opened 2018-06-14
David White

kauaicrusier:


danitacheny:
Great work! The site looks great, your code looks organized, everything looks good.

One suggestion I have is to make more frequent commits with more meaningful messages whenever you make a significant change to your code. This makes it easier for you and anyone else who reads your code to understand what you've done.

Keep up the great work!

PR #396, opened 2018-06-14
Adrian Bracamonte

adrianbraca:


danitacheny:
Fantastic job! Everything looks well organized and the site looks great! The only issue I see is that the layout gets a little wonky at widths between 768px and 875px.

Good job making commits for different stages of your work! I would definitely recommend doing it even more frequently, with meaningful messages whenever you make a significant change in your code.

Overall, nice job!

PR #408, opened 2018-07-31
Updated HTML and CSS

verydecent:
Introduced global font percentage to HTML in the CSS file

font-size: 62.5%;

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -21,7 +21,7 @@ time, mark, audio, video {
 	border: 0;
 	font: inherit;
     vertical-align: baseline;
-    font-size: 62.5%;
+    font-size: 100%;
 }
 /* HTML5 display-role reset for older browsers */
 article, aside, details, figcaption, figure, 
@@ -55,6 +55,7 @@ table {
 html, body {
     height: 100%;
     font-family: 'Titillium Web', sans-serif;
+    font-size: 62.5%;
 }
 
 h1, h2, h3, h4, h5 {
@@ -74,8 +75,8 @@ h1, h2, h3, h4, h5 {
 header {
     width: 100%;
     display: flex;
-    justify-content: space-around;
-    align-items: flex-end;
+    justify-content: space-between;
+
 }
 
 header nav {
@@ -83,11 +84,13 @@ header nav {
     width: 69.3%;
     justify-content: space-between;
     display: flex;
+    font-size: 1.6rem;
 }
 
 header nav a {
     text-decoration: none;
     color: #b0b0b0;
+
 }
 
 header nav a:hover {
@@ -108,7 +111,7 @@ header .logo {
 }
 
 .cta .cta-text {
-    font-size: 72px;
+    font-size: 7.2rem;
     text-align: center;
     letter-spacing: 5px;
     display: flex;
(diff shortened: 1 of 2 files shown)


gooseandmegander:
Thank you for the PR.

css/index.css
```diff
@@ -1 +1,301 @@
...
+
+.cta .cta-text {
+    font-size: 7.2rem;
+    text-align: center;
+    letter-spacing: 5px;
+    display: flex;
+    justify-content: center;
+    flex-flow: column;
+    align-items: center;
+    margin: 0 9.31% 0 0;
```
gooseandmegander, line 121: Nicely done % conversions.

css/index.css
```diff
@@ -1 +1,301 @@
...
+    margin-top: 40px;
+    margin-bottom: 20px;
+}
+
+/* Tablet */
+@media (max-width: 800px) {
+
+    .container {
+
+        max-width: 800px;
```
gooseandmegander, line 190: The max-width property is already set by the media query. `width: 100%` might be what you meant.

css/index.css
```diff
@@ -1 +1,301 @@
...
+    }
+  
+    header nav {
+        width: 96%;
+        padding: 0 3.75%;
+    }
+
+    header .logo {
+        order: -1;
+        margin: 30px 100% 0;
```
gooseandmegander, line 204: It looks like you accidentally wrote 100% in the `margin: 30px 100% 0;` line. You could change the 100% to auto to fix that problem.

css/index.css
```diff
@@ -1 +1,301 @@
...
+    .contact {
+        margin: 0 3.75%;
+    }
+}
+
+/* Phone */
+@media (max-width: 500px) {
+
+    .container {
+        max-width: 500px;
```
gooseandmegander, line 240: Same as above with the tablet media query, the max-width is already set.

css/index.css
```diff
@@ -1 +1,301 @@
...
+    }
+
+    header .logo {
+        margin-top: 20px;
+        margin-bottom: 20px;
+    }
+
+    .cta {
+        margin-top: 50px;
+        flex-flow: column;
```
gooseandmegander, line 274: Why are you using flex-flow and align-items here?

css/index.css
```diff
@@ -1 +1,301 @@
...
+        align-items: center;
+    }
+
+    .cta img {
```
gooseandmegander, line 278: This block isn't necessary since the img has `display: none` on it in the tablet view. The effect will cascade.

css/index.css
```diff
@@ -1 +1,301 @@
...
+.main-content .bottom-content {
+    display: flex;
+}
+
+.main-content .bottom-content p {
+    padding-right: 20px 2.5%;
+}
+
+/* Contact Styles */
+
```
gooseandmegander, line 170: You can add something like this to your contact styles:

>.contact h4 {
  font-size: 1.6rem;
}

to get the Contact header to be the right size.

gooseandmegander (commented):
Really nice job, Wonjae. You clearly understand %-based layout techniques. I have a few comments for you to review.

PR #419, opened 2018-07-31
Responsive Web Design II - Evan Carlstrom

ecarlstrom:
Project completed, no horizontal scroll bar at any window or viewport size.

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -60,12 +60,12 @@ html, body {
 
 h1, h2, h3, h4, h5 {
     font-family: 'Bangers', cursive;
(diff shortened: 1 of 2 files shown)


crypto-jones:
Overall, the code looks clean & is easy to read. It looks like you got into some of the stretch goals, which is great! My only critique is minor, but definitely important. Be careful with the consistency of you spacing in the CSS, like if you use 2 spaces, stick with it throughout the project. This is especially important as we move into JS to make your code as readable as possible because things can't get messy pretty quickly.

PR #423, opened 2018-07-31
RWD II Assignment

ckirkwood94:


gooseandmegander:
Thank you for the PR.

css/index.css
```diff
@@ -1 +1,492 @@
...
+}
+
+h1, h2, h3, h4, h5 {
+    font-family: 'Bangers', cursive;
+    letter-spacing: 1px;
+    margin-bottom: 15px;
+}
+
+.container {
+    width: 880px;
```
gooseandmegander, line 67: You've already declared the max-width as 880px. To make the container responsive and %-based, `width: 100%`.

css/index.css
```diff
@@ -1 +1,492 @@
...
+   font-size: 1.6rem;
+   line-height: 18px;
+}
+
+.main-content .top-content {
+    display: flex;
+}
+
+.main-content .middle-img {
+    display: block;
```
gooseandmegander, line 155: If you add a percent-based width constraint to the .middle-img, it will get rid of the horizontal scrollbar. Do you remember why?

css/index.css
```diff
@@ -1 +1,492 @@
...
+    
+    .cta .cta-text {
+        margin: 0;
+    }
+
+    .cta img {
+        display: none;
+    }
+
+    .main-content .middle-img {
```
gooseandmegander, line 219: You don't need to have his block, if you add the width property in your desktop view, `.main-content .middle-img { ... }` since the width should already be in effect.

css/index.css
```diff
@@ -1 +1,492 @@
...
+        background: gray;
+    }
+
+    header .logo {
+        margin: 20px 0;
+    }
+
+    .cta {
+        margin-top: 50px;
+        flex-flow: column;
```
gooseandmegander, line 274: You could omit flex-flow and align-items. You only have one item, and the elements within it are already aligned.

css/index.css
```diff
@@ -1 +1,492 @@
...
+        align-items: center;
+    }
+
+    .cta img {
```
gooseandmegander, line 278: The selector block `.cta img { ... }` isn't affecting the layout in mobile since the img has the property `display: none` in effect in tablet view. It cascades down to effect the mobile view as well.

css/index.css
```diff
@@ -1 +1,492 @@
...
+
+    .contact {
+        margin: 20px 4%;
+    }
+}
+
+/* Services page CSS specific styling start below*/
+
+.services-header-img	{
+	width: 100%;
```
gooseandmegander, line 310: Nice job catching the `width: 100%;` on the img.

css/index.css
```diff
@@ -1 +1,492 @@
...
+    .text-content {
+        margin: 0 1% 0 1.6%;
+    }
+
+    .contact {
+        margin: 20px 4%;
+    }
+}
+
+/* Services page CSS specific styling start below*/
```
gooseandmegander, line 307: You have the services class as sibling to you container class. You should be using it for all of your Services page selectors.

gooseandmegander (commented):
Great job, Caleb. The services page is bomb! The only tweak I'd make there is styling on the services-mobile-buttons. I've left a few notes for you to review.

PR #426, opened 2018-07-31
Tristen Neu -- RWD-II

ghost:


crypto-jones:
Overall, your code is clean & easy to read. Looks like you got into the stretch goals, which is awesome. Great job! My only critique is consider committing your changes more often & list them by the task you complete vs. just pushing up basic goals as a commit & stretch as another. The value of committing often will be seen later on once we get into bigger projects & it's especially important when you work on a team with others. Not a big deal, but definitely a great habit to get into. Keep up the great work!

PR #435, opened 2018-07-31
Updated Document -Candace Syeinebach

CandaceStein:
I converted units to rems and percentage per instructions.


phillipsdev:
Thank you for the PR! 👍

PR #441, opened 2018-07-31
changing to percent! MT

TrevMike:


css/index.css
```diff
@@ -1 +1,464 @@
...
+h2,
+h3,
+h4,
+h5 {
+  font-family: 'Bangers', cursive;
+  letter-spacing: 1px;
+  margin-bottom: 1.7%;
+}
+
+.container {
```
zackhitch, line 150: Perfect set up for the container class with max-width and width with percentages.

css/index.css
```diff
@@ -1 +1,464 @@
...
+
+/* Header Styles */
+
+header {
+  width: 100%;
+  display: flex;
+}
+
+header nav {
+  margin-top: 4.54%;
```
zackhitch, line 164: Percentages for left and right margins are a must, but not necessarily so for the vertical margins.

css/index.css
```diff
@@ -1 +1,464 @@
...
+  width: 69.31%;
+  justify-content: space-between;
+  display: flex;
+}
+
+header nav a {
+  text-decoration: none;
+  color: #b0b0b0;
```
zackhitch, line 172: Great job using hex color values and rem units for font-sizes.

zackhitch (approved):
Really amazing job here! Code is clean and easy to read. Your layout beautifully matches the design files, and you don't repeat code, in keeping with the cascading principle. Keep it up!

PR #442, opened 2018-07-31
Jcassidy - Updated to be responsive

cassidyjamesw:


css/index.css
```diff
@@ -1 +1,383 @@
...
+
+/* Set every element's box-sizing to border-box */
+* {
+    box-sizing: border-box;
+}
+
+html, body {
+    height: 100%;
+    font-family: 'Titillium Web', sans-serif;
+    font-size: 62.5%;
```
zackhitch, line 60: Add a selector for _just_ `html { font-size: 62.5%}` above line 57, as setting that as the font size here has unintended styling consequences.

css/index.css
```diff
@@ -1 +1,383 @@
...
+	justify-content: flex-start;
+}
+
+
+
+
+/* TABLET VIEW */
+@media (max-width: 800px) {
+    .container {
+        width: 100%;
```
zackhitch, line 276: Great job remembering to set container width to 100%.

css/index.css
```diff
@@ -1 +1,383 @@
...
+    }
+}
+
+@media (max-width: 500px) {
+    .container {
+        width: 100%;
+    }
+    .container header {
+
+        display: flex;
```
zackhitch, line 311: Remember the cascading principle of CSS, there is no need to repeat yourself with `display: flex` here, as it was set to that earlier (higher) in the code.

index.html
```diff
@@ -3,19 +3,88 @@
...
 
   <link href="https://fonts.googleapis.com/css?family=Bangers|Titillium+Web" rel="stylesheet">
   <link rel="stylesheet" href="css/index.css">
 
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  
   <!--[if lt IE 9]>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/html5shiv/3.7.3/html5shiv.js"></script>
   <![endif]-->
 </head>
```
zackhitch, line 17: You need to have `<body>` tags in your html file. While it's recommended to have a containing `div` that wraps all other content, wrapping _that_ div should always be a `body` tag

css/index.css
```diff
@@ -1 +1,383 @@
...
+    width: 100%;
+}
+
+.cta .cta-text {
+	width: 100%;
+    display: flex;
+    flex-direction: column;
+    align-items: flex-start;
+    flex-wrap: wrap;
+    font-size: 77px;
```
zackhitch, line 133: Try changing font-sizes from px to rem

css/index.css
```diff
@@ -1 +1,383 @@
...
+  width: 50%;
+  margin-right: 4%;
+}
+
+.main-content .top-content .text-content:last-child {
+  margin-right: 0;
+}
+
+.main-content .middle-img {
+  margin-top: 5%;
```
zackhitch, line 181: Left and Right margins definitely need to be changed to percents, but not necessarily for the vertical margins.

zackhitch (approved):
Good job with this project! The font-sizes are a bit off, check where you set the font-size to 62.5%. They layout matches nicely when one line of code is changed. 

All in all, good job with percentage based layout!

PR #454, opened 2018-07-31
Linda Yang - Responsive Web Design II stretch

lyang9:
Still working on making it pixel perfect


lyang9:
Not pixel close yet

crypto-jones:
Overall, the code looks clean & is easy to read. It looks like you got into some of the stretch goals, which is great. Awesome job! My only critique is minor, but definitely important. Be careful with the consistency of you spacing in the CSS, like if you use 2 spaces, stick with it throughout the project. This is especially important as we move into JS to make your code as readable as possible because things can't get messy pretty quickly.

PR #458, opened 2018-07-31
Andrew Ghobrial - Responsive-Web-Design-II

ghobs91:
@BigKnell @gooseandmegander

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -1,1 +1,434 @@
-/* Use your CSS from RWD - I here.  IF you don't have CSS, use the solution CSS from RWD - I*/
+/* http://meyerweb.com/eric/tools/css/reset/ 
+   v2.0 | 20110126
+   License: none (public domain)
+*/
+
+html, body, div, span, applet, object, iframe,
+h1, h2, h3, h4, h5, h6, p, blockquote, pre,
+a, abbr, acronym, address, big, cite, code,
+del, dfn, em, img, ins, kbd, q, s, samp,
+small, strike, strong, sub, sup, tt, var,
+b, u, i, center,
+dl, dt, dd, ol, ul, li,
+fieldset, form, label, legend,
+table, caption, tbody, tfoot, thead, tr, th, td,
+article, aside, canvas, details, embed, 
+figure, figcaption, footer, header, hgroup, 
+menu, nav, output, ruby, section, summary,
+time, mark, audio, video {
+	margin: 0;
+	padding: 0;
+	border: 0;
+	font-size: 100%;
+	font: inherit;
+	vertical-align: baseline;
+}
+/* HTML5 display-role reset for older browsers */
+article, aside, details, figcaption, figure, 
+footer, header, hgroup, menu, nav, section {
+	display: block;
+}
+body {
+	line-height: 1;
+}
+ol, ul {
+	list-style: none;
+}
+blockquote, q {
+	quotes: none;
+}
+blockquote:before, blockquote:after,
+q:before, q:after {
+	content: '';
+	content: none;
+}
+table {
+	border-collapse: collapse;
+	border-spacing: 0;
+}
+
+/* Set every element's box-sizing to border-box */
+* {
+    box-sizing: border-box;
+}
+
+html {
+    /* 10px if the browser was 16px*/
+     font-size: 62.5%; 
+    
+  /*   1 rem === 10px */
+    
+  }
+
+html, body {
+    height: 100%;
+    font-family: 'Titillium Web', sans-serif;
+}
+
+h1, h2, h3, h4, h5 {
+    font-family: 'Bangers', cursive;
+    letter-spacing: 1px;
+    margin-bottom: 15px;
+}
+
+.container {
+  width: 880px;
+  /* TB LR */
+  margin: 0 auto;
+}
+
+/* header */
+
+header {
+	display: flex;
+	flex-direction: row;
+	justify-content: space-evenly;
+	flex-wrap: wrap;
+	align-content: flex-start;
+}
+
+header nav {
+  display: flex;
+  width: 80%;
+  justify-content: space-evenly;   
+  margin-top: 35px;
+}
+
+header nav a {
+  /* display: inline-block; */
+  /* margin-right: 65px; */
+  font-size: 1.077rem;
+  color: gray;
+  text-decoration: none;
+}
+
+header nav a:hover {
+  text-decoration: underline;
+}
+
+header nav a:first-child {
+  margin-left: 0;
+}
+
+header nav a:last-child {
+  margin-right: 0;
+}
+
+header .logo {
+	width: 20%;
+	margin-top: 25px;
+}
+
+/* cta */
+
+.cta {
+	display: flex;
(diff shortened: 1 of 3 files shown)


gooseandmegander:
Thank you for the PR.

gooseandmegander:
Thank you for the PR.

css/index.css
```diff
@@ -1 +1,434 @@
...
+
+html, body {
+    height: 100%;
+    font-family: 'Titillium Web', sans-serif;
+}
+
+h1, h2, h3, h4, h5 {
+    font-family: 'Bangers', cursive;
+    letter-spacing: 1px;
+    margin-bottom: 15px;
```
gooseandmegander, line 71: You can put a font-size property in this block to create a basic heading font-size.

gooseandmegander: If you go the route of adding the font-size in the the headings block, you can also make a `p { ... }` selector block to do the same for your p elements. Then you have a standard from the get-go and only need to make changes for special circumstances (e.g. "Innovation On Demand").

css/index.css
```diff
@@ -1 +1,434 @@
...
+}
+
+.container {
+  width: 880px;
```
gooseandmegander, line 75: Adding `width: 100%;` and `max-width: 880px;` will give the .container responsiveness. This was the cause of the horizontal scrollbar.

css/index.css
```diff
@@ -1 +1,434 @@
...
+  /* TB LR */
+  margin: 0 auto;
+}
+
+/* header */
+
+header {
+	display: flex;
```
gooseandmegander, line 83: There are some properties in this block that are unnecessary. Which ones are they and why?

css/index.css
```diff
@@ -1 +1,434 @@
...
+  display: flex;
+  width: 80%;
+  justify-content: space-evenly;   
+  margin-top: 35px;
+}
+
+header nav a {
+  /* display: inline-block; */
+  /* margin-right: 65px; */
+  font-size: 1.077rem;
```
gooseandmegander, line 100: I would bump this font-size up.

css/index.css
```diff
@@ -1 +1,434 @@
...
+	flex-direction: row;
+	justify-content: space-evenly;
+	flex-wrap: wrap;
+	align-content: flex-start;
+}
+
+header nav {
+  display: flex;
+  width: 80%;
+  justify-content: space-evenly;
```
gooseandmegander, line 93: `space-evenly` isn't widely cross-browser compatible, and another choice that would get you closer to the design file is `space-between`. It looks like your `header nav a:first-child { ... }` and `header nav a:last-child { ... }` aren't actually doing what they should be.

(Part I of space-between implementation)

css/index.css
```diff
@@ -1 +1,434 @@
...
+header nav a:first-child {
+  margin-left: 0;
+}
+
+header nav a:last-child {
+  margin-right: 0;
+}
+
+header .logo {
+	width: 20%;
```
gooseandmegander, line 118: No need for the width property here.

css/index.css
```diff
@@ -1 +1,434 @@
...
+	margin-top: 25px;
```
gooseandmegander, line 119: You can change this to a margin property like `margin: 25px 0 0 ?` to get that space between the logo and the nav.
This is if you implement the space-between I mentioned above.

(Part III of space-between implementation)

css/index.css
```diff
@@ -1 +1,434 @@
...
+  font-size: 1.077rem;
+  color: gray;
+  text-decoration: none;
+}
+
+header nav a:hover {
+  text-decoration: underline;
+}
+
+header nav a:first-child {
```
gooseandmegander, line 109: If you implement the space-between as mentioned above, you can change the margin-left here to get some space between (pun not intended) the services link and the left side of the viewport. Otherwise, if you want a margin on the whole header nav, you can add `margin: 0 1%;` to the `header { ... }` block. Then you'll have some white space between the left viewport side and services link && the logo and right viewport side.

If any of these notes are not logical, DM me. I'd be happy to show you what I mean.

(Part II of space-between implementation)

css/index.css
```diff
@@ -1 +1,434 @@
...
+header .logo {
+	width: 20%;
+	margin-top: 25px;
+}
+
+/* cta */
+
+.cta {
+	display: flex;
+	flex-direction: row;
```
gooseandmegander, line 126: `flex-direction: row` is not necessary. The row flex direction is default with `display:flex`.

gooseandmegander: Why are you using `flex-wrap: wrap;` and `align-content: flex-start`?

css/index.css
```diff
@@ -1 +1,434 @@
...
+	flex-wrap: wrap;
+	align-content: flex-start;
+}
+
+.cta .cta-text {
+  display: flex;
+  flex-direction: column;
+  align-content: center;
+  vertical-align: top;
+  font-size: 5.9rem;
```
gooseandmegander, line 137: If you put the font-size in the headings block I mentioned above, then you will need to extract the font-size for "Innovation On Demand" out to its own selector block, which may be a good idea (pun not intended) anyway.

css/index.css
```diff
@@ -1 +1,434 @@
...
+  margin-top: 120px;
+  text-align: center;
+  /* padding: 0px 75px; */
+}
+
+.cta .cta-text button {
+  background: white;
+  vertical-align: top;
```
gooseandmegander, line 145: What is `vertical-align:top` doing?

gooseandmegander: You'll also probably want a width on the button.

css/index.css
```diff
@@ -1 +1,434 @@
...
+	flex-direction: row;
+	justify-content: space-evenly;
+	flex-wrap: wrap;
+	align-content: flex-start;
+}
+
+.cta .cta-text {
+  display: flex;
+  flex-direction: column;
+  align-content: center;
```
gooseandmegander, line 135: Consider align-items rather than align-content.

css/index.css
```diff
@@ -1 +1,434 @@
...
+	display: flex;
+	flex-direction: column;
+	border-top: 1px solid black;
+	margin-top: 76px;
+	padding-top: 5%;
+}
+
+.main-content .top-content {
+	display: flex;
+	flex-direction: row;
```
gooseandmegander, line 174: `flex-direction: row;` and `justify-content: space-evenly;` are showing up a lot. What is their purpose?

css/index.css
```diff
@@ -1 +1,434 @@
...
+    header {
+        width:100%;
+        display: flex;
+        flex-direction: column-reverse;
+        justify-content: space-evenly;
+        flex-wrap: wrap;
+        align-items: center;
+    }
+
+    header nav {
```
gooseandmegander, line 330: The nav needs some work. If you want some idea, feel free to DM me.

gooseandmegander (commented):
Great job, Andrew. The first thing I immediately notice is the font-sizes are a bit small. I've got some comments for you to review and some suggestions for those font-sizes and the nav in desktop view. It looks great and the services page is awesome.

PR #459, opened 2018-07-31
Max Kajiwara - Responsive-Web-Design-II

maxkajiwara:
Paired programming with Michael Agard


crypto-jones:
Overall, the code looks clean & is easy to read. Awesome job! I know you pair programmed with someone, but just for the sake of the critique, which is minor — Be careful with the consistency of your spacing in the CSS, like if you use 2 spaces, stick with it throughout the project. This is especially important as we move into JS to make your code as readable as possible because things can get messy pretty quickly. Other than that, I have nothing else. You're doing an amazing job. Keep it up!

PR #461, opened 2018-07-31
Tom Kisner Responsive Design 2 Project

ThomasKisner:


index.html
```diff
@@ -1,21 +1,106 @@
...
+      <a href="products.html">Product</a>
+      <a href="">Vision</a>
+      <a href="">Features</a>
+      <a href="about.html">About</a>
+      <a href="">Contact</a>
+      <img class="logo" src="img/logo.png" alt="Great Idea! Company logo.">
+    </nav>
+
+
+    <header>
```
zackhitch, line 32: Convention would be to wrap the nav and in `header`, and to use a `section` tag here instead.

index.html
```diff
@@ -1,21 +1,106 @@
...
+        <div class="vision">
+          <h3>Vision</h3>
+          <p>Aliquam elementum magna eros, ac posuere elvit tempus et. Suspendisse vel tempus odio, in interdutm nisi. Suspendisse
+            eu ornare nisl. Nullam convallis augue justo, at imperdiet metus scelerisque quis.
+          </p>
+        </div>
+      </section>
+
+      <footer>
+        <contact>
```
zackhitch, line 91: A `section` with a class of `contact` would be more appropriate here. This will work because of XML conventions and the tags match, but `<contact>` is not an HTML element.

about.html
```diff
@@ -0,0 +1,84 @@
...
+</section>
+
+<section class="work-together">   
+  <h2> Let's Work Together</h2>
+    <p>Leverage agile frameworks to provide a robust synopsis for high level overviews. Iterative approaches to corporate strategy foster collaborative thinking to further the overall value proposition. Organically grow the holistic world view of disruptive innovation via workplace diversity and empowerment.</p>
+    </section>
+
+        
+    <footer>
+      <contact>
```
zackhitch, line 71: As in index.html, `contact` is not a valid HTML element. Try a `section` with contact as the class.

about.html
```diff
@@ -0,0 +1,84 @@
...
+    <nav>
+        <img  src="img/lambda-black.png" alt="lambda logo" class="services-img">
+        <a href="index.html">Home</a>
+        <a href="about.html">About</a>
+        <a href="">Products</a>
+        <a href="">Blog</a>
+        <a href="contact.html">Contact</a>
+    </nav>
+
+    <header class="header-about-page">
```
zackhitch, line 25: Same with index.html regarding header vs section here.

css/index.css
```diff
@@ -1 +1,856 @@
...
+
+
+/* Copy and paste your work from yesterday here and start to refactor into flexbox */
+
+/* http://meyerweb.com/eric/tools/css/reset/ 
+   v2.0 | 20110126
+   License: none (public domain)
+*/
+
+html, body, div, span, applet, object, iframe,
```
zackhitch, line 76: Looks like the reset was applied twice! A problem with copy and pasting that happens to everyone at least once! Be sure to check what you're copy/pasting!

css/index.css
```diff
@@ -1 +1,856 @@
...
+
+h1, h2, h3, h4, h5 {
+    font-family: 'Bangers', cursive;
+    letter-spacing: 1px;
+    margin-bottom: 15px;
+}
+
+/* Your code starts here! */
+
+body .container{
```
zackhitch, line 144: Convention would be to leave `body` off this selector, so just `.container {...}` is appropriate. (The level of specificity can be confusing, I know). 

Great job with the max-width, I also would recommend setting width here to 100%.

css/index.css
```diff
@@ -1 +1,856 @@
...
+nav{
+	display: flex;
+	height: 63px;
+	justify-content: space-between;
+	align-items: flex-end;
+
+}
+
+nav a {
+	font-size: 1.5rem;
```
zackhitch, line 158: Great use of percentages and rem units!

css/index.css
```diff
@@ -1 +1,856 @@
...
+    padding: 10px 0;
+}
+
+nav img {
+    margin: 30px 0;
+}
+
+header {
+    flex-direction: column;
+    height:417px;
```
zackhitch, line 581: Don't forget a space after the `:`, its a good convention to follow, as it makes everything easier to read.

zackhitch (approved):
Really well done! Your code is really clean, there are a few minor syntax items that could be cleaned up, but you're on track for sure! 

Outstanding job pushing yourself into the stretch goals!

PR #465, opened 2018-08-01
Initial submission.

keirankozlowski:
Still needs some work. I need help with something about responsive, it's not sizing properly still.

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -1,1 +1,289 @@
-/* Use your CSS from RWD - I here.  IF you don't have CSS, use the solution CSS from RWD - I*/
+/* http://meyerweb.com/eric/tools/css/reset/ 
+   v2.0 | 20110126
+   License: none (public domain)
+*/
+
+html, body, div, span, applet, object, iframe,
+h1, h2, h3, h4, h5, h6, p, blockquote, pre,
+a, abbr, acronym, address, big, cite, code,
+del, dfn, em, img, ins, kbd, q, s, samp,
+small, strike, strong, sub, sup, tt, var,
+b, u, i, center,
+dl, dt, dd, ol, ul, li,
+fieldset, form, label, legend,
+table, caption, tbody, tfoot, thead, tr, th, td,
+article, aside, canvas, details, embed, 
+figure, figcaption, footer, header, hgroup, 
+menu, nav, output, ruby, section, summary,
+time, mark, audio, video {
+	margin: 0;
+	padding: 0;
+	border: 0;
+	font-size: 100%;
+	font: inherit;
+	vertical-align: baseline;
+}
+/* HTML5 display-role reset for older browsers */
+article, aside, details, figcaption, figure, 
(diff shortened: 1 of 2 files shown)


gooseandmegander:
Thank you for the PR.

css/index.css
```diff
@@ -1 +1,347 @@
...
+}
+
+p {
+	font-size: 1.6rem;
+}
+
+.container {
+    max-width: 880px;
+	width: 100%;
+	margin-left: 5.68%;
```
gooseandmegander, line 78: Setting the `margin-left` and `margin-right` properties are the culprits to your horizontal scrollbar. Taking those out and adding `margin: auto;` will fix the issue.

css/index.css
```diff
@@ -1 +1,347 @@
...
+/* Header Styles */
+
+header {
+    width: 100%;
+    display: flex;
+	font-size: 1.6rem;
+}
+
+header nav {
+    width: 69.32%;
```
gooseandmegander, line 91: With the .container fix above, the nav width will need to be changed to `width: 100%`. Otherwise, it won't fill the entire nav container.

css/index.css
```diff
@@ -1 +1,347 @@
...
+
+header .logo {
+    margin: 3.18% 0 0 10.34%;
+}
+
+/* Call To Action (CTA) Styles */
+
+.cta {
+    display: flex;
+    margin-top: 8.18%;
```
gooseandmegander, line 115: Be careful using % with margin/padding top or bottom. We don't have a hard bound on the vertical aspect of elements usually, so the percentages won't have anything to use as a base.

css/index.css
```diff
@@ -1 +1,347 @@
...
+    text-align: center;
+    margin: 4.55% 0 2.27%;
+}
+
+/* Tablet */
+@media (max-width: 800px) {
+
+    .container {
+        width: 100%;
+		max-width: 800px;
```
gooseandmegander, line 212: `max-width: 800px;` is unnecessary here. The media query is already taking care of the max-width.

css/index.css
```diff
@@ -1 +1,347 @@
...
+
+    .contact {
+        margin: 0 3.75%;
+    }
+}
+
+/* Phone */
+@media (max-width: 500px) {
+
+    .container {
```
gooseandmegander, line 272: The .container block in this media query is unnecessary.

css/index.css
```diff
@@ -1 +1,347 @@
...
+        flex-wrap: wrap;
+        align-content: space-between;
+        height: 88%;
+    }
+
+	.main-content .bottom-content .text-content {
+		margin-top: 7.5%;
+	}
+	
+    .text-content {
```
gooseandmegander, line 340: You can leave this one if you want, but I don't see that much of a difference with it there.

gooseandmegander (commented):
Awesome, Keiran. I don't have many notes for you, but there are a few to review.

PR #466, opened 2018-08-01
Lavon RWD2

micah-shrak:
RWD2 updates

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -1,1 +1,293 @@
 /* Use your CSS from RWD - I here.  IF you don't have CSS, use the solution CSS from RWD - I*/
+/* http://meyerweb.com/eric/tools/css/reset/ 
+   v2.0 | 20110126
+   License: none (public domain)
+*/
+
+html, body, div, span, applet, object, iframe,
+h1, h2, h3, h4, h5, h6, p, blockquote, pre,
+a, abbr, acronym, address, big, cite, code,
+del, dfn, em, img, ins, kbd, q, s, samp,
+small, strike, strong, sub, sup, tt, var,
+b, u, i, center,
+dl, dt, dd, ol, ul, li,
+fieldset, form, label, legend,
+table, caption, tbody, tfoot, thead, tr, th, td,
+article, aside, canvas, details, embed, 
+figure, figcaption, footer, header, hgroup, 
+menu, nav, output, ruby, section, summary,
+time, mark, audio, video {
+	margin: 0;
+	padding: 0;
+	border: 0;
+	font-size: 100%;
+	font: inherit;
+	vertical-align: baseline;
+}
+/* HTML5 display-role reset for older browsers */
+article, aside, details, figcaption, figure, 
+footer, header, hgroup, menu, nav, section {
+	display: block;
+}
+body {
+	line-height: 1;
+}
+ol, ul {
+	list-style: none;
+}
+blockquote, q {
+	quotes: none;
+}
+blockquote:before, blockquote:after,
+q:before, q:after {
+	content: '';
+	content: none;
+}
+table {
+	border-collapse: collapse;
+	border-spacing: 0;
+}
+
+/* Set every element's box-sizing to border-box */
+* {
+    box-sizing: border-box;
+}
+
+html, body {
+    height: 100%;
+    font-size: 62.5%;
+    font-family: 'Titillium Web', sans-serif;
+}
+
+h1, h2, h3, h4, h5 {
+    font-family: 'Bangers', cursive;
+    letter-spacing: 1px;
+    margin-bottom: 15px;
+}
+
+.container {
+    max-width: 880px;
+    margin: auto;
+}
+
+/* Header Styles */
+
+header {
+    width: 100%;
+    display: flex;
+}
+
+header nav {
+    margin-top: 4.545%; 
+    width: 69.3181%;
+    justify-content: space-between;
+    display: flex;
+}
+
+header nav a {
+    text-decoration: none;
+    color: #b0b0b0;
+}
+
+header nav a:hover {
+    text-decoration: underline;
+    color: black;
+}
+
+header .logo {
+    margin: 2.8% 0 0 9.1%;
+}
+
+/* Call To Action (CTA) Styles */
+
+.cta {
+    display: flex;
+    margin-top: 8.18181%;
+    justify-content: flex-end;
+}
+
+.cta .cta-text {
+    font-size: 4.5rem;
+    text-align: center;
+    letter-spacing: 5px;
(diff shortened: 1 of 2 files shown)


gooseandmegander:
Thank you for the PR.

css/index.css
```diff
@@ -1 +1,293 @@
...
+    height: 100%;
+    font-size: 62.5%;
+    font-family: 'Titillium Web', sans-serif;
+}
+
+h1, h2, h3, h4, h5 {
+    font-family: 'Bangers', cursive;
+    letter-spacing: 1px;
+    margin-bottom: 15px;
+}
```
gooseandmegander, line 66: Not sure what happened to the `p { ... }` block, but setting a font-size of 1.6rem will make your text larger.

css/index.css
```diff
@@ -1 +1,293 @@
...
+
+.container {
+    max-width: 880px;
```
gooseandmegander, line 69: Missing the `width: 100%;` to make the .container percent-based responsive.

css/index.css
```diff
@@ -1 +1,293 @@
...
+    margin: auto;
+}
+
+/* Header Styles */
+
+header {
+    width: 100%;
+    display: flex;
```
gooseandmegander, line 77: Adding `font-size: ?rem;` here will make your nav links larger and closer to the design file.

css/index.css
```diff
@@ -1 +1,293 @@
...
+}
+
+header nav {
+    margin-top: 4.545%;
```
gooseandmegander, line 81: Be wary of using %s for top and bottom spacing utilities. There isn't a hard bound on the vertical aspect of elements usually, so the percentages won't have anything to use as a base.

Also, you could change this to a margin property to add in some left and right margin. For instance, `margin: 4.545% 1% 0;` will give a top margin of 4.545%, left and right margin of 1%, and a bottom margin of 0. This will give the page some 'breathing room' and the nav won't seem like it's pushed against the viewport.

css/index.css
```diff
@@ -1 +1,293 @@
...
+    font-size: 4.5rem;
+    text-align: center;
+    letter-spacing: 5px;
+    display: flex;
+    justify-content: center;
+    flex-flow: column;
+    align-items: center;
+    margin: 0 8.2% 0 0;
+}
+
```
gooseandmegander, line 119: Here would be a good place to add the `.cta .cta-text h1 { ... }` selector to change the font-size of the `h1`. You can then delete line 110. You'll also find the rem unit will need to be larger than 4.5.

css/index.css
```diff
@@ -1 +1,293 @@
...
+    box-sizing: border-box;
+}
+
+html, body {
+    height: 100%;
+    font-size: 62.5%;
+    font-family: 'Titillium Web', sans-serif;
+}
+
+h1, h2, h3, h4, h5 {
```
gooseandmegander, line 62: In this block, you can add a font-size for your `h1`-`h6` elements. The headings in the .text-content areas are a bit small. Although, depending on how you're styling the "Innovation On Demand", you may need to specify its font-size in a new selector block with `.cta .cta-text h1 { ... }` to override the set font-size here.

css/index.css
```diff
@@ -1 +1,293 @@
...
+    display: flex;
+    justify-content: center;
+    flex-flow: column;
+    align-items: center;
+    margin: 0 8.2% 0 0;
+}
+
+.cta .cta-text button {
+    border: 1px double black;
+    font-size: 1rem;
```
gooseandmegander, line 122: The text on the button could be a little larger. Maybe somewhere between 1.4rem and 1.8rem?

css/index.css
```diff
@@ -1 +1,293 @@
...
+   font-size: 1rem;
+   line-height: 18px;
+}
+
+.main-content .top-content {
+    display: flex;
+}
+
+.main-content .middle-img {
+    display: block;
```
gooseandmegander, line 151: If you add `width: 100%;` here, it will get rid of the horizontal scrollbars. Do you remember why?

css/index.css
```diff
@@ -1 +1,293 @@
...
+
+/* Tablet */
+@media (max-width: 800px) {
+
+    .container {
+        width: 100%;
+    }
+
+    header {
+        display: flex;
```
gooseandmegander, line 184: No need for `display: flex;` here. It's already in your header selector block for the desktop view, so it is already in effect in the tablet view.

css/index.css
```diff
@@ -1 +1,293 @@
...
+
+/* Phone */
+@media (max-width: 500px) {
+
+    header {
+        justify-content: center;
+    }
+
+    header nav {
+        width: 100%;
```
gooseandmegander, line 237: You could have the nav width be 100% from the desktop view, and it wouldn't change your browser view. If you did that, you would save 2 lines of code (1 in each media query).

css/index.css
```diff
@@ -1 +1,293 @@
...
+        margin: 0;
+        padding: 0;
+    }
+
+    header nav a {
+        width: 100%;
+        display: block;
+        text-align: center;
+        padding: 2.0% 0;
+        font-size: 1.25rem;
```
gooseandmegander, line 248: The font-size could be bumped up for the mobile view. Also, there could be more spacing between the nav links and the logo.

css/index.css
```diff
@@ -1 +1,293 @@
...
+        background: gray;
+    }
+
+    header .logo {
+        margin: 2.0% 0;
+    }
+
+    .cta {
+        margin-top: 10%;
+        flex-flow: column;
```
gooseandmegander, line 266: I think you could omit the flex-flow and align-items property in mobile. They're not showing any significant changes in the browser.

css/index.css
```diff
@@ -1 +1,293 @@
...
+        align-items: center;
+    }
+
+    .cta img {
```
gooseandmegander, line 270: This block, `.cta img { ... }` isn't necessary since it's has `display: none` in effect in the tablet view and therefore, the mobile view as well.

css/index.css
```diff
@@ -1 +1,293 @@
...
+
+    .main-content .top-content {
+        justify-content: space-around;
+        flex-direction: column;
+        height: 30%;
+    }
+
+    .main-content .bottom-content {
+        flex-wrap: wrap;
+        align-content: space-between;
```
gooseandmegander, line 282: You could also omit this align-content property.

gooseandmegander (commented):
Great job on RWD-II, Lavon. I have some comments for you to review. They're mostly small things here and there. You did great utilizing the %-based layout technique!

One thing to keep in mind is to commit often. It gives you a decent safety net as these projects get bigger.

PR #467, opened 2018-08-02
Louis Li - Responsive-Web-Design-II

louislee7:
@kiadorii

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -1,1 +1,368 @@
 /* Use your CSS from RWD - I here.  IF you don't have CSS, use the solution CSS from RWD - I*/
+
+/* http://meyerweb.com/eric/tools/css/reset/ 
+   v2.0 | 20110126
+   License: none (public domain)
+*/
+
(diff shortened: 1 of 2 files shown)


css/index.css
```diff
@@ -1 +1,368 @@
...
+    width: 32rem;
+}
+
+hr {
+	width: 100%;
+}
+
+.p2	{
+	width: 100%;
+	margin: 2rem;
```
kiachoi, line 159: I would recommend using % for margins and rem for font sizes.

kiachoi (approved):
✔️ What Went Well:
----
+ Introduced font-size: 62.5%; into your html selector.

🎗 Recommendation(s):
----
+ The alignment for the phone view is slightly off - the right is smaller than the left.
+ For tablet view - I would recommend spacing out the nav bar a little lower. More of a personal preference.

Other:
----
+ Nice work Louis! If you are interested going further on this project, see how close to you can get it to be pixel perfect! 😄

PR #472, opened 2018-08-02
Abi Franklin RWD-II

AbiFranklin:
@gakko1


gakko1:
## Great

Impressed by the BEM usage! It looks correct from looking it over as well, very nice Abi! Concise CSS, looks good! Liking the scrolling center image as well as the adaptive image and logo.

## Requested Improvements

A suggestion for the "Innovation On Demand" and button to be responsive in size along with the image and logo (looks great though!). BEM makes it so much more readable, my only other suggestion would be to divide your CSS into sections with comments which will help on large CSS files in the future.

## Rating: {1-3}

3

PR #474, opened 2018-08-02
Colin Dismuke - Responsive Web Design II

cpdis:
Beginning to work on the rest of the CSS to make the site fully responsive. I had already used percentages so the transitions between sizes are pretty smooth already. 

@John-Spraul


cpdis:
The working site can be found here: [Responsive Web Design II](https://responsive-ii.netlify.com/)

John-Spraul:
Good commits!
Good specific commit messages!
Thanks for including your name and the assignment in the PR title!

I see a horizontal scrollbar on `index.html` at 892px to 801px 🙀 😱 (if you can't figure out how to fix it, let me know and we'll figure it out)
class names `paragraph-one` through `-four` are never referenced in the CSS files
services.html seems organized pretty well 👍 
instructions say `Constrain your container element with a max-width of 880px`; you used `width: 880px` instead
ask me about my margin/padding question, when you have time (we'll discuss some stuff) ❕ 


As long as you can fix that horizontal scroll bar, then it's all good! 👍 
Nice work on the stretch goals! 💯 🎉 
Make sure you follow the instructions before ✔️ing the item on the list

cpdis:
Fixed the horizontal scroll bar by adding `width: 100%` to the `.container` class. Updated `width: 880px` to `max-width: 880px`. Removed the unused `paragraph` classes.

Thanks for the feedback 🙏  I'll make sure I carefully read each instruction before checking them off next time, haha.

PR #480, opened 2018-08-02
Responsive Web Design II

ghost:
I was told that I should open my pull request as soon as possible so that you can easy see the changes I am making while working on the project. I hope I am doing this correctly.

@kiadorii

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -102,6 +102,9 @@ body {
 
 .nav-text {
     color: rgb(186,186,186);
+    font-size: 1rem;
+    font-weight: bold;
+    
 }
 
 #logo-holder {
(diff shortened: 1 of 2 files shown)


css/index.css
```diff
@@ -1 +1,583 @@
...
+	justify-content: space-between;
+    align-items: center;
+}
+
+.title-text {
+	font-size: 1.125rem;
+}
+
+.title-text2 {
+	font-size: 1.125rem;
```
kiachoi, line 192: By setting font-size to 62.5%, you can get rid of the extra decimals you have here.

kiachoi (approved):
✔️ What Went Well:
----

🎗 Recommendation(s):
----
+ I know you mentioned about having to make font-size 62.5% and I would still recommend it. Especially since it was mentioned in the project.

Other:
----
+ Always impressed with what you're able to put out.

PR #481, opened 2018-08-02
Wesley Meeks- Initial Commit

ghost:
Added HTML and CSS from previous project and added font-size:62.5% to HTML selector

@kiadorii

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -54,7 +54,7 @@ table {
 html{
   font-size: 62.5%;
 }
-html, body {
+.container {
     height: 100%;
     font-family: 'Titillium Web', sans-serif;
 		max-width: 880px;
@@ -72,8 +72,8 @@ h1, h2, h3, h4, h5 {
 }
 
 body{
-	margin-left: 10px;
-	margin-right: 10px;
+	margin-left: 1.14%;
+	margin-right: 1.14%;
 }
 .nav{
 	display: flex;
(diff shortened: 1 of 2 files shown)


css/index.css
```diff
@@ -72,8 +72,8 @@ h1, h2, h3, h4, h5 {
 }
 
 body{
-	margin-left: 10px;
-	margin-right: 10px;
+	margin-left: 1.14%;
```
kiachoi: If you're thinking of refactoring this, it can be condensed to be `margin: 0 1.14%`

css/index.css
```diff
@@ -105,18 +105,17 @@ a{
...
 
 }
-
 .button{
 	display: flex;
 	justify-content: center;
-	margin-left: 60px;
-	padding: 5px 35px 5px 35px;
+	margin-left: 22%;
+	padding: .9% 10% .9% 10%;
```
kiachoi: It's more common to see top and bottom to be in px, and left and right to be in percentages.

kiachoi (approved):
✔️ What Went Well:
----
+ Introduced font-size: 62.5%; into your html selector.

🎗 Recommendation(s):
----
+ There were some spacing issues throughout the views but I can see the foundation of it. Just keep messing around with and working on it!

Other:
----
+ Let me know if you have any questions in doing something. I definitely don't want you to spend your whole night struggling. But on that note, you're making progress!

PR #490, opened 2018-08-02
Henry Moland - Added html and css style sheet to project

henrymoland:
Added my html and css style sheet to project
push files to Github


47analogy:
Nice work! There isn't a horizontal scrollbar. I especially like how you used an animation at 800px and 500px. Only thing I can say is try to better align your code in the css  for better readability (or consider an editor extension that does it for you when saving)

PR #491, opened 2018-08-02
Took previous lesson, and made it more responsive with percentages, m…

JJHutton81:
…ax-widths, and rem font size.

Finished basic assignment. Will come back for stretch goals time permitting.

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -226,7 +226,7 @@ section{
 }
 
 .middle-img {
-    margin-top: 31px;
+    margin: 30px 0px 30px 0px;
     border: 1px solid black;
     border-radius: 15px;
     height: 182px;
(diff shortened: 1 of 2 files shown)


John-Spraul:
More commits!
More specific commit messages!
Include your name and the assignment in the PR title!

It's so beautiful and responsive! 💯 

Nav bar goes a git wonky at ~650px.
I get a horizontal scroll bar on the mobile view (500px), which I was able to fix by tinkering around with the stylings that included `left` and `top` and `position: relative`
All left/right margins should be in %'s (lines 96, 337, 349 of `index.css`)
All `font-size` should be in `rem` instead of `px` (line 255 of `index.css`)

Good work matching the design files at each break point. 👍 
Try to be more thorough, use `Ctrl+F` to search through your code quickly 🔍 
You rely on `top` and `left` WAY too much. You should be able to get the styling you want without using these. (trust the flex 💪)

PR #499, opened 2018-08-02
[Tommy Collison] initial commit, updated font

tommycollison:
Ping @John-Spraul.

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -54,7 +54,7 @@ table {
 html, body {
     height: 100%;
     font-family: 'Titillium Web', cursive;
-
+	font-size: 62.5%; /* font-size 1em = 10px on default browser settings */ 
 }
 
 h1, h2, h3, h4, h5 {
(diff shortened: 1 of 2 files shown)


tommycollison:
Getting started now.

John-Spraul:
More commits!
Good specific commit messages! 🎉 
Good job including name, but also include the assignment in the PR title!

Why isn't this finished, Tommy? 😭 

Your use of id's makes you repeat code:
```
#bottom_left_text {
	/* 880 / 3 is ~290. Let's go with 280. */
/*	max-width: 280px;
*/	width: 30%;
	display: inline-flex;
}

#bottom_center_text {
/*	max-width: 280px;
*/	width: 30%;
	display: inline-flex;

}

#bottom_right_text { 
	/*max-width: 280px;*/
	width: 30%;
	display: inline-flex;
}
```
All left/right widths, margins, and padding should be in `%` instead of `px` (line 85, 94, 120, 121 in `index.css`)
`font-size`s should be in `rem` instead of `px`

Did you not read the instructions? Did you not have time to finish? Please let me know.

tommycollison:
@John-Spraul Erg, I think I ran out of time and then moved onto Preprocessors. I think I didn't grok that all right/left margins, etc, can be %, not just the texts. I think it'd be useful to go back and practice targeting specific elements and not repeating by fixing/redoing (and, y'know, finishing) this assignment. I'll get it done before tomorrow night. Thanks!

PR #500, opened 2018-08-02
Responsive-Web-Design-II/

Dewayne87:
WORK IN PROGRESS

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -64,6 +64,10 @@ h1, h2, h3, h4, h5 {
 }
 
 /* Your code starts here! */
+@keyframes example {
+    from {background-color:lightskyblue;}
+    to {background-color:gold;}
+}
 #main-container{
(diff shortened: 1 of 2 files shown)


John-Spraul:
Good # of commits!
Could be more specific with your commit messages!
Good job including the assignment in the PR title; please include your name as well!

Ooo, fancy animation! 😮 

CSS formatting could use some work. Should look something like (notice the spacing):
```
.class1 {
    property: value;
}

.class2 {
     other-property: value;
}
etc.
```
I'd argue that you use too many id's, but I am biased 🤔
good job choosing helpful id's and class names
`Max-width` should be `max-width` (line 73 of index.css)

Good job on this assignment! 👍 
Don't be afraid to add your own personal touch to assignments in the future 🍨 🦅 🌰 🌮

PR #505, opened 2018-08-02
rewritten css of Great Idea, with rem and % based widths

chillingtonbillington:
I was able to clean up my CSS up so much with this. Got rid of any redundant or useless code as well!

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -1,1 +1,256 @@
 /* Use your CSS from RWD - I here.  IF you don't have CSS, use the solution CSS from RWD - I*/
+
+/* Use your own code or past solution for Great Idea Web Page CSS here! */
+
+
+/* http://meyerweb.com/eric/tools/css/reset/ 
+   v2.0 | 20110126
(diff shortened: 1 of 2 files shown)


John-Spraul:
More commits!
Good specific commit message!
Please include your name and the assignment (`Responsive-Web-Design-II` in this case) in the PR title!

It's so responsive and beautiful 😭
Some class names and id's could be more specific (`#gs`)? 🤷‍♂️ 
I always argue that classes are better than id's, but I'd like to hear your opinion on when each should be used.

Good job focusing on get rid of redundancies 👍 
Have you found a consistent fix for fixing images stretching all funkily? ❓
Generally, CSS should be structured like this:
```
main {
	max-width:880px;
	margin: 0 auto;
}

header {
    display:flex;
    justify-content:space-between;
	
}
/* new line */
.classname /* space */ {
    property: value;
}
```

Good job on this assignment. Focus on following good git practices in the future.

PR #510, opened 2018-08-03
Oliver Abreu - Responsive Web Design: The Sequel

AbreuOliver:
@FireInjun


decagondev (commented):
## Overview
Oliver. The overall code is clean and readable. Your desktop css could do with a little work and your tablet css is not too bad..

## Breakdown
### Good Points
- you have created a fully responsive design
- your measurements are in percentages and relatives
- You have leveraged off the flexbox class

### Things to consider
- try to utilize space around instead of space between in the navbar and logo area
- consider using max-width property for the container and give width 90%
- think of other ways that flexbox could make your life easier in this project

## Overview
Oliver your use of flexbox is an interesting well formed choice. you may need to think about reading up on more of the possible settings and attributes in the flexbox arsenal to leverage the best use of it to style a truly responsive design. also try to be mindful of git granularity along with git regularity for commits.

PR #512, opened 2018-08-03
Brandon Moll RWD II

BrandonMoll:
@gakko1

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -233,13 +233,14 @@ header nav a {
 */
 
 .contact {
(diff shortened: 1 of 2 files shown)


gakko1:
## Great

Love the transitions for break point resizing. Don't necessarily need the webkit prefixes for older browsers but it's good practice. Matches design files, looks good.  

## Requested Improvements

Maybe adding BEM or another page for extra practice given time. 

## Rating: {1-3}

3

PR #514, opened 2018-08-03
Kendra Williams - Responsive Design II - index and services

codingforkicksOG:
@John-Spraul


John-Spraul:
Good number of commits! 🔢 
Commit messages could be more specific, but this is a good improvement! 🎉 
Great job including your name and the assignment in the PR title 🎊 🎆 

It's so beautiful and responsive 😻 

Keep your class names consistent. (eg. `middle-img` vs `middletext`)
I don't like `navi` as a class name, but maybe I'm biased? 🤷‍♂️
Consider separating class names with hyphens or underscores (`mid-text-bottom` instead of `midtextbottom`)

Great work on this assignment, especially with the stretch goals 👍
Keep up the good work!

PR #525, opened 2018-08-03
(Lorenzo Evans) Responsive Design Layout

LorenzoEvans:
Completed the Great Idea website, entirely refactored for responsive design, still looking to improve the structuring of the CSS, prune some unnecessary code.


47analogy:
Nice job! There isn't a horizontal scroll bar. Only suggestion is to maybe give some more space on top the h3 tag at 500px

PR #528, opened 2018-08-03
responsive navbars for all sizes, using percentages and rems complete…

jbierly1:
…d...not pixel perfect but on the way


jbierly1:
@juliankohlman

PR #536, opened 2018-08-03
Mitchell Robles - Respsonsive Web Design II

mitchellr92:
Remade page from scratch for extra practice and better functionality.

Put in break points with media queries.

Used percentages for all margins, padding and widths.

Changed font sizes using rem.

@John-Spraul

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -60,6 +60,10 @@ h1, h2, h3, h4, h5 {
 
 /* Use your CSS from RWD - I here.  IF you don't have CSS, use the solution CSS from RWD - I*/
 
+html {
+    font-size: 62.5%;
+}
+
 body {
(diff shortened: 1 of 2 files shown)


John-Spraul:
Good number of commit messages! 🔢 
Great specific commit messages! 🎉 
FANTASTIC job of including your name and the assignment in the PR title 🕺 🎈 

It's so beautifully responsive! 😻 

Is `<container>` a real HTML element? doesn't seem to break anything, so 🤷‍♂️ 
Is it necessary to repeat this so many times?:
```
body {
    width: 90%;
    max-width: ___px;
    margin: 0 auto;
}
```

Great work on this assignment! Feel free to add your own personal touch to these assignments.
Keep up the good work 👍

PR #538, opened 2018-08-03
Chris Juelfs Responsive Web Design II complete!

CheapCharlie:
I committed at multiple stages like requested
I still haven't went back to case-correct my classes and id's like you recommended but i will keep that in mind in future assignments that start from scratch.
Hopefully ill have time in the near future to attempt to make it pixel-perfect!


47analogy:
Everything looks good, no horizontal scroll bar. Just a litlle tidying up of html and css for readability.

PR #542, opened 2018-08-04
Marc de la Torre CSPT2 Responsive Web Design II new pull

Mdelatorre1314:
@camilamdaniels


camilamdaniels:
Great job! Thanks for taking the time to update the files so that the css would render properly :)

PR #543, opened 2018-08-04
Ben Masterson-Finished Responsive-Web-Design-II

bennybooftw:


John-Spraul:
More commits!
Commit messages are better, but could be more specific! 👍
Great inclusion of your name and the assignment in the PR title 💯

It's so responsive and beautiful 😭 

Did you skip this step? `Introduce font-size: 62.5%; into your html selector.`
No confusing class names or id's that I can see 👀 👏

Keep up the good work! Try to make a habit of committing small changes often.

PR #545, opened 2018-08-07
Jerrard Smith- Responsive Design II

JerrardSmithOfficial:
@47analogy


47analogy:
No html content currently. See that it's working locally, so will revisit when code is pushed up to remote.

PR #556, opened 2018-09-04
Trevor Lewis

Trevor-Lewis:
@Wdifulvio523


css/index.css
```diff
@@ -1 +1,646 @@
...
+    width: 100%;
+	margin: 10px 0px 30px 0px;
+}
+
+
+.Contact {
+	width: 15.9%;
+}
+
+footer {
```
Wdifulvio523, line 181: the footers in the index and serviced pages are different. Is this intended?

Wdifulvio523:
Great job on this project Trevor! Both the Index and Services pages look great. My only note is that some things should be more uniform between the two pages, such as the container size or the footer. As usual, well done on this!

PR #557, opened 2018-09-04
Responsive-Web_Design-II Conner H

TRIF3X:
pull request for project  Responsive-Web_Design-II

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -427,21 +427,21 @@ h1, h2, h3, h4, h5 {
         display: flex;
         flex-direction: column-reverse;
         height: 8rem;
+        width: 100%;
     }
     .container header .logo {
-        margin-left: 24.5rem;
+        align-self: center;
         margin-top: .3rem;
-        width: 17%;
     }
     .container header nav {
         font-size: 1.25rem;
         display: flex;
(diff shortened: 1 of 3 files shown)


css/index.css
```diff
@@ -1 +1,711 @@
...
+        height: 20rem;
+        margin-bottom: 8.2rem;
+    }
+    .container header .logo {
+        padding-bottom: 2rem;
+    }
+    .container header nav {
+        flex-direction: column;
+        text-align: center;
+        height: 20rem;
```
Wdifulvio523, line 668: if you look on devtools, this is kind of getting pushed out due to the a-tags.

css/index.css
```diff
@@ -1 +1,711 @@
...
+    }
+    .container header nav a {
+        padding: 1rem 0rem;
```
Wdifulvio523, line 671: This padding is what's pushing the a tags outside of it's parent container. How can we get the desired effect without the negative side-effect?

Wdifulvio523:
Very good job on this project Conner! Your index page looks fantastic. I just had one note on the nav items in mobile-view, which I commented. As usual, well done on this! Keep it up.

PR #562, opened 2018-09-04
RileyB-RWD-2

Riley-Brown:
Copied files over


css/index.css
```diff
@@ -1 +1,388 @@
...
+
+/* nav styles */
+
+html {
+  font-size: 62.5%;
+}
+body {
+  animation: fadein 2s;
+}
+@keyframes fadein {
```
Wdifulvio523, line 12: Awesome animation

Wdifulvio523:
Very well done on this project Riley! Both the index and services pages are fantastic. Even better, they are consistent with each other (things like the container and footer are the same on both pages). Good job on this, keep it up!

PR #563, opened 2018-09-04
Zachary Arney

zacharyarney:


luiscmartinez (commented):
Well done, all of MVP met. Conversions done properly. No horizontal scroll bar. Would like to see more detailed commits on what you done to your code. Instead of just completed, give more detail on what you completed & so on

PR #566, opened 2018-09-04
Responsive Web Design 2

David247365:
@CookieMonsta89


CookieMonsta89:
Great job on your project.  The only thing that I notice is that you introduced 62.5% font-size to the html but you never changed the font size in your nav so the text is really small.  I would do 1.6rem font size on the nav.  Good job man.  Keep up the good work

PR #568, opened 2018-09-04
standard assignment complete

bkwillia052:
@kamry-bowman Nav issue from RWD I assignment resolved.

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -1,1 +1,379 @@
-/* Use your CSS from RWD - I here.  IF you don't have CSS, use the solution CSS from RWD - I*/
+/* http://meyerweb.com/eric/tools/css/reset/ 
+   v2.0 | 20110126
+   License: none (public domain)
+*/
+
+html, body, div, span, applet, object, iframe,
+h1, h2, h3, h4, h5, h6, p, blockquote, pre,
+a, abbr, acronym, address, big, cite, code,
+del, dfn, em, img, ins, kbd, q, s, samp,
+small, strike, strong, sub, sup, tt, var,
+b, u, i, center,
+dl, dt, dd, ol, ul, li,
+fieldset, form, label, legend,
+table, caption, tbody, tfoot, thead, tr, th, td,
+article, aside, canvas, details, embed, 
+figure, figcaption, footer, header, hgroup, 
+menu, nav, output, ruby, section, summary,
+time, mark, audio, video {
+	margin: 0;
+	padding: 0;
+	border: 0;
+	font-size: 100%;
+	font: inherit;
+	vertical-align: baseline;
+}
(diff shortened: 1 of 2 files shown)


css/index.css
```diff
@@ -1 +1,379 @@
...
+	body .inner-body {
+		display: flex;
+		flex-direction: column;
+	}
+}
+
+body {
+	background: rgba(5, 159, 255, .6);
+	margin: auto;
+	overflow: hidden;
```
kamry-bowman, line 143: don't do this! it hides all content on your page longer than one screen height. I could only see half of your page and had  until I removed this property. In general, you shouldn't be using overflow: hidden unless you have a good reason that the content is not fitting into its container properly.

css/index.css
```diff
@@ -1 +1,379 @@
...
+
+/* Copy and paste your work from yesterday here and start to refactor into flexbox */
+* {
+	box-sizing: border-box;
+}
+
+html {
+	font-size: 62.5%;
+}
+
```
kamry-bowman, line 74: remember, the pattern is html { font-size: 62.5%}, then body { font-size: 1.6rem }. That way every item within the body will default to a basic font-size of 16px, and you can adjust headings as you want them

css/index.css
```diff
@@ -1 +1,379 @@
...
+.services .body-container {
+	width: 60%;
+	margin: 0 auto;
+	background: white;
+}
+header {
+	width: 100%;
+	display: flex;
+	height: auto;
+	position: relative;
```
kamry-bowman, line 161: position relative is used when you want to shift an element out of its normal position, or you want to you position: absolute on its child element. neither apply here, so I don't think you need it.

css/index.css
```diff
@@ -1 +1,379 @@
...
+	margin-bottom: 15px;
+	justify-content: space-between;
+}
+
+header hr {
+	display: none;
```
kamry-bowman, line 167: you probably don't need this element at all if you're not displaying it

css/index.css
```diff
@@ -1 +1,379 @@
...
+	font-size: 1.6rem;
+	padding: 10px 1rem;
+	width: 75%;
+}
+
+
+.inner-body {
+display: flex;
+flex-direction: row;
+margin: 10px 1rem;
```
kamry-bowman, line 264: remember, when using rem on box model, it will change when the user changes their default font-size. make sure you want that behavior. often times you won't

kamry-bowman (commented):
Nice work, you really have got the percentage, responsive, and fluid concepts down! Site looks good at all levels. I would recommend going back over this with Pixel Perfect if you have time this week. It can be eye opening how different our sites are from the designs when we've only tried to eyeball it.

PR #574, opened 2018-09-04
Responsive Design II

CLantigua2:
Carlos Lantigua

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -126,7 +126,7 @@ h1, h2, h3, h4, h5 {
 }
 
 .container {
-  width: 880px;
+  max-width: 880px;
   /* TB LR */
   margin: 0 auto;
 }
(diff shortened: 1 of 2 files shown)


Velsu (commented):
Great job on the assignment Carlos. You successfully completed the MVP by making website responsive, this time utilizing percentages and other responsive units. 

Navbar is alligns properly in tablet and mobile view thanks to properly implemented flexbox rules, although i believe it's width should be adjusted a bit during transition from tablet to mobile as it is spanning too wide.

The one thing that potentially needs improving is transition from desktop mobile views. There is visible horizontal line when decreasing the resolution as well as some different behavior of the elements on the page. That code could be tweaked a little to make transition a little smoother and overall better experience.

Your code is clean and You have certainly improved and made Your code a lot more DRY without repeating unecessary rules.
Your git workflow also improved and is a lot more fluid now, solid job 👍

PR #575, opened 2018-09-04
Responsive-Web-Design-II Blake Koennecker

koennecker:


Velsu (commented):
Excellent job Blake, nothing really to add much. Just remember You do not need and should not repeat same rules again in media queries because they are already applied at the root, for example display: flex.

PR #576, opened 2018-09-04
Jake Thomas - Responsive-Web-Design-II

jmaxt12:
Jake Thomas

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -20,12 +20,13 @@ article, aside, canvas, details, embed,
 figure, figcaption, footer, header, hgroup, 
 menu, nav, output, ruby, section, summary,
 time, mark, audio, video {
-	margin: 0;
+	margin: 0 auto;
 	padding: 0;
 	border: 0;
 	font-size: 100%;
 	font: inherit;
-	vertical-align: baseline;
+    vertical-align: baseline;
+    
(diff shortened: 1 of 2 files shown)


css/index.css
```diff
@@ -1 +1,451 @@
...
+    font-size: 1rem;
+}
+
+}
+
+
+/* Mobile View */
+@media (max-width:500px) {
+
+  .cta img{
```
Velsu, line 388: Yet again, no reason to repeat that rule. It is enoug if it was declared once at the top and it will take effect unless explicitly specified.

Velsu (commented):
Great job on the assignment Jake. You successfully completed the MVP by making website responsive, this time utilizing percentages and other responsive units.

Navbar is alligns properly in tablet view and mobile view thanks to properly implemented flexbox rules,

The one thing that potentially needs improving is transition from desktop mobile views. There is visible horizontal line when decreasing the resolution as well as some different behavior of the elements on the page. That code could be tweaked a little to make transition a little smoother by centering navbar and getting rid of horizontal scrollbar by adjusting width of body element.

Your code is could be a little more clean by not repeating unecessary rules and adhering to DRY principles.

Solid job 👍 

PR #577, opened 2018-09-04
McKay Bonham - Responsive Web Design II

MEBonham:


jeffyjkang:
awesome stuff, i would consider adding some margins/padding on the smaller media queries but other than that, amazing and smooth functionality!

PR #579, opened 2018-09-04
Troy Amlee [sprint] completed base objectives

troyAmlee:
Attempting Stretch


cesarnml:
Excellent job making index.html fully responsive. Neat use animations on hover and on media-breaks. services.html looks horrible, but that was a stretch ... Good job! Good commit history.

PR #582, opened 2018-09-04
Lola Heffernan - RWD II

lolax:


css/index.css
```diff
@@ -1 +1,496 @@
...
+}
+
+.btm-text {
+	margin-right: 45px;
+}
+
+ul {
+	margin: 25px 0;
+	line-height: 2;
+}
```
Wdifulvio523, line 278: minor sizing issues on mobile/tablet

css/index.css
```diff
@@ -1 +1,496 @@
...
+
+.about-full p {
+  font-size: 18px;
+}
+
+.about-text {
+  border: 1px solid grey;
+  border-radius: 2px;
+  width: 750px;
```
Wdifulvio523, line 317: minor sizing issues here caused by the hard-coded width

Wdifulvio523:
Absolutely fantastic job on this project Lola! The index, services, and about pages all look fantastic. Even more, they are all consistent with each other - things like the nav are uniform on all three pages, that's the way you wanna be! Some sizing/space issues on the services and about page which I commented, but overall it looks great! Keep up the good work!

PR #583, opened 2018-09-04
Katie Fitzpatrick -Responsive-Web-Design-II

KatieFizzy:


css/index.css
```diff
@@ -1 +1,819 @@
...
+    flex-direction:row;
+   
+
+}
+
+#navBar nav a {
+	display: flex;
+	flex-direction: row;
+	justify-content: space-between;
+	margin: 5%;
```
kamry-bowman, line 106: remember that if you use margin 5% that also sets the margin of the top and bottom to 5%. If you don't have a specified height on its container (nav), then if anything changes the height of nav, it will also change the top and bottom margins of your anchor tags. That is not necessarily bad, but it could be surprising behavior that is hard to diagnose (like if you add something tall to nav, your anchor tags start moving around too for some reason)

kamry-bowman (commented):
Awesome job on desktop, and tablet pretty much pixel perfect! Nice progress on the mobile, just need to figure out the horizontal alignment on the hero-text, the breaking on the Get Started button, and the vertical alignment. If you find yourself with some extra time before the Friday Sprint Challenge, that could be a great area to work on.

As a general note, there's not so much of a benefit to using percentages on vertical items (like margin-top and margin-bottom for example), since responsiveness isn't a huge factor vertically speaking. It can create surprising behavior sometimes as well, and makes it harder to do precise adjustments.

PR #587, opened 2018-09-04
RWD II

brianmgre:
Brian Gregorius

@CookieMonsta89


CookieMonsta89:
Good job on this assignment.  Everything seems to be responsive when I use the viewport.  One thing I noticed is that you did a 10px font size in your html and then in sections, you just did a 100% font size.  Great idea and it does work so no need to use rem.

PR #588, opened 2018-09-04
RWD 2

edjeong99:


luiscmartinez (commented):
Conversions from px to rems or percentage are well done. Implemented good breakpoints as instructions implied. Keep it up.
You got some horizontal scroll bars because some elements overflow outside your container

PR #589, opened 2018-09-04
Daniel Ferrer, Responsive Web Design II

ghost:


jeffyjkang:
looks good same thing as project 1

PR #590, opened 2018-09-04
RWD 2 Pull Request

ethiapath:
@CookieMonsta89


CookieMonsta89:
This looks pretty good.  You have responsive elements which is what we are asking for.  One recommendation, when you added 62.5% font-size to your html, it shrinks your text.  You then have to go through and add 1.6rem to the font in order for it to be 16pixels = 1.6rem.

PR #596, opened 2018-09-04
YanrongXu - Responsive-Web-Design-II

YanrongXu:
@kamry-bowman

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -243,4 +243,39 @@ footer {
 }
 
 @media (max-width: 500px){
-   
+    
+    .container header nav {
+        flex-direction: column;
+        border-top: 1px solid silver;
+    }
+
+    .container header nav a {
+        padding: 20px 0;
+        margin: 0;
+        text-align: center;
+        border-bottom: 1px solid silver;
+    }
+
+    .cta {
+        margin: 40px 0;
+    }
+
+    .main-content {
+        margin: 50px 0 40px 0;
+        font-size: 1.7rem;
+    }
+
+    .main-content .top-content {
+        flex-direction: column;
+    }
+
+    .main-content .text-content {
+        margin: 20px 0;
+        padding: 0;
+    }
+
+    .main-content .bottom-content {
+        flex-direction: column;
+    }
+
+}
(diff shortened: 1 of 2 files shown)


css/index.css
```diff
@@ -1 +1,281 @@
...
+h1, h2, h3, h4, h5 {
+    font-family: 'Bangers', cursive;
+    letter-spacing: 1px;
+    margin-bottom: 15px;
+}
+
+.container {
+    max-width: 880px;
+    width: 100% ;
+    padding: 2%;
```
kamry-bowman, line 74: if you set the max-width to 880px, but then keep 2% padding on the left and right, your inner content will only ever get as large as 96% of 880px, or 844px. That's why your desktop view doesn't quite match up on Pixel Perfect

css/index.css
```diff
@@ -1 +1,281 @@
...
+
+html {
+    font-size: 62.5%;
+}
+
+html, body {
+    height: 100%;
+    font-family: 'Titillium Web', sans-serif;
+}
+
```
kamry-bowman, line 64: when you use html { font-size: 62.5% }, I would encourage you to set body { font-size: 1.6rem }, then everything defaults to 16px while being responsive. You then only have to set your h1-h6 custom

css/index.css
```diff
@@ -1 +1,281 @@
...
+footer {
+    text-align: center;
+    margin: 40px 0 20px;
+}
+
+@media (max-width: 800px) {
+    
+    .container {
+        width: 98%;
```
kamry-bowman, line 185: this stops anything from reaching all the way across, which you don't want. Your buttons, for example, should go to the edges of the page. Keep your width at 100%, and if you need something to be set back from the edge, use margins or padding further down in your nesting structure

css/index.css
```diff
@@ -1 +1,281 @@
...
+    .container header nav a {
+        font-size: 1.7rem;
+    }
+
+    .container header .logo {
+        display: flex;
+        max-width: 100%;
+        height: auto;
+        justify-items: center;
+        margin: 0 0;
```
kamry-bowman, line 208: you can set margin: 0

kamry-bowman (commented):
Everything looks great, I'm very happy with this! Pretty dialed in on the Pixel Perfect, with relatively small deviations. I left some notes on specific items I saw in your code

PR #600, opened 2018-09-04
Responsive-Web-Design-II - By Michael Checo

MichaelCheco:


KevinBrack:
Looking good sir. The nav text looks really tiny until you get to mobile. Also when in mobile I think the nav margin is a little bigger than in the design file. Other wise you are killing it with the responsive design. Keep it up.

PR #601, opened 2018-09-04
Responsive-Web-Design-II by Katy Hollobaugh

KHollobaugh:


KevinBrack:
This is really fantastic. It matches the design files so well! One little pointer, I noticed the margin on the main container shifting everything to the right a little when on mobile view. If you set the container margin to 0 auto, and put that distance in the padding instead, you should still get that nice boundary but it should help with centering. Let me know if you need any help. Thank you!

PR #607, opened 2018-09-04
Hunter Smith - RWD Project 2

Hunter315:
@kamry-bowman

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -1,1 +1,386 @@
-/* Use your CSS from RWD - I here.  IF you don't have CSS, use the solution CSS from RWD - I*/
+/* http://meyerweb.com/eric/tools/css/reset/ 
+   v2.0 | 20110126
+   License: none (public domain)
+*/
+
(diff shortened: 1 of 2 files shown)


Velsu (commented):
Great job on the assignment Hunter. You successfully completed the MVP by making website responsive, this time utilizing percentages and other responsive units.

Navbar is alligns properly in tablet and mobile view thanks to properly implemented flexbox rules.

The one thing that potentially needs improving is transition from desktop mobile views. There is visible horizontal line when decreasing the resolution as well as some different behavior of the elements on the page. That code could be tweaked a little to make the banner picture have width in percentages which would most likely solve the issue of horizontal scrollbar on mobile breakpoint.

I also believe the tablet vieww needs some tweaking. As soon as we lower the width from desktop to tablet view the big round image disappears, but the text gets pushed a little to the side and it does not look too user friendly.

Your code is clean. Your git workflow could be a little more organized and descriptive.

PR #609, opened 2018-09-04
Edwin Torres

ETorresM2:
Edwin Torres

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -1,1 +1,255 @@
-/* Use your CSS from RWD - I here.  IF you don't have CSS, use the solution CSS from RWD - I*/
+/* http://meyerweb.com/eric/tools/css/reset/ 
+   v2.0 | 20110126
+   License: none (public domain)
+*/
+
+html, body, div, span, applet, object, iframe,
+h1, h2, h3, h4, h5, h6, p, blockquote, pre,
+a, abbr, acronym, address, big, cite, code,
+del, dfn, em, img, ins, kbd, q, s, samp,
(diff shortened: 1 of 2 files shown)


css/index.css
```diff
@@ -1 +1,255 @@
...
+}
+
+h1, h2, h3, h4, h5 {
+    font-family: 'Bangers', cursive;
+    letter-spacing: 1px;
+    margin-bottom: 15px;
+}
+
+html {
+  font-size: 62.5%;
```
Velsu, line 67: Good job converting the font size to percentages.

css/index.css
```diff
@@ -1 +1,255 @@
...
+  display: flex;
+  width: 100%;
+  margin-top: 35px;
+  justify-content: flex-start;
+  margin-left: 5%
+}
+
+header nav a {
+  display: flex;
+  margin-right: 5%;
```
Velsu, line 94: That margin will cause issues on the mobile view. It will create horizontal scroll bar so i would advise to remove it when that breakpoint is reached.

Velsu (commented):
Great job Edwin. You successfully converted all fixed width's to either percentages or rems. Your website is fully responsive, although not without small issues. Your footer has too small font and that needs adjusting. There is also an issue with horizontal scroll bar in mobile view due to incorrect margins, please check comments below for more details.

Your git workflow requires improvement, You need to do more frequent and more descriptive commit messages. With each functionality make a new commit, for example for each breakpoint, separate these into different commit categories etc.

Overall You did good job 👍 

PR #610, opened 2018-09-04
Kyran McCann

kyranmccann:


luiscmartinez:
Great, usage of commits. can see what changes you made. & met requirement of changing pixels into percentage.

PR #612, opened 2018-09-04
Responsive Design for tablet and mobile

Bryan-GF:
@jeffyjkang


jeffyjkang:
great stuff, i think i only see one issue, in the desktop view the contact and great idea logo.

PR #613, opened 2018-09-04
Responsive-Web-Design-II - By Yusuf Nafey

yusufnafey:


KevinBrack:
This is flawless work sir. Matches the layout well, but beyond that the elements flow and adjust well to every resolution. Keep it up!!

PR #615, opened 2018-09-04
RWD2

codypeak:
ran out of time


KevinBrack:
Excellent work sir. Really showing you have the hang of responsive design. The nav links font is showing up tiny on my machine, and when on mobile view the link borders are not extending to the edges yet. Other than that it is looking fantastic! If you need any help let me know.

PR #620, opened 2018-09-05
Responsive-Web-Design-II - By Edd Burke

bummings:
Couple of conflicts in the media queries- but yknow ;)


KevinBrack:
Getting there. I believe the CTA heading is quite a bit larger than the design file. Seems to be throwing that section off quite a bit. Looks like you know what you are doing though. Let me know if you need help

PR #624, opened 2018-09-05
Finished Project

Drewsup123:
Sorry this is late must have forgot


css/index.css
```diff
@@ -1 +1,318 @@
...
+
+h1, h2, h3, h4, h5 {
+    font-family: 'Bangers', cursive;
+    letter-spacing: 1px;
+    margin-bottom: 15px;
+}
+h2{
+    font-size:1.6rem;
+}
+p{
```
kamry-bowman, line 68: could do this as p, ul, ol, a { to be more general

kamry-bowman (commented):
Looks awesome! Destkop is perfect! Tablet is right on, except for some slight deviations on the header section. Mobile is also great, just some vertical issues from Pixel Perfect. Great!

PR #625, opened 2018-09-05
John Graniero

JohnGran66:


css/index.css
```diff
@@ -1 +1,310 @@
...
+   .contact {
+       margin: 0 30px;
+   }
+}
+
+
+
+@media (max-width: 500px) {
+  
+  }
```
Wdifulvio523, line 245: you need to delete this closing bracket so that everything else below this is nested inside the media query. Otherwise, it affects Desktop view and tablet view.

css/index.css
```diff
@@ -1 +1,310 @@
...
+      font-size: 20px;
+      border-bottom: 1px solid gray;
+  }
+  
+  header nav a:first-child {
+      border-top: 1px solid gray;
+  }
+
+  header nav a:hover {
+      background: gray;
```
Wdifulvio523, line 275: try to avoid making the background the same color as the font, it will make the font invisible.

css/index.css
```diff
@@ -1 +1,310 @@
...
+      justify-content: space-around;
+      flex-direction: column;
+      height: 300px;
+      padding-bottom: 28px;
+  }
+
+  .main-content .bottom-content {
+      flex-wrap: wrap;
+      align-content: space-between;
+      height: 440px;
```
Wdifulvio523, line 298: this hard-coded height is causing a lot of whitespace on the page, is that intended? Try removing the height in DevTools, see how it looks!

Wdifulvio523:
Pretty good job on this project John! You had an extra closing bracket in your mobile-view, which was causing the mobile-view to show up on tablet and desktop-views. I commented on it so you can clear it up quickly. Keep working and we'll get back on track in no time! Please reach out if you have any questions.

PR #626, opened 2018-10-03
Christene Bennett 

christenebennett:
@abravebee Adaptive Website project completed.


abravebee:
* You’ve knocked it out of the park again! I love to see that you’re pushing yourself on the stretch goals as well and it sounded in stand-up like you had a lot of fun with that, which is great to hear. Your code organization is also extremely neat and compact. My only complaint is that you’re making my job too easy. :) Excellent work all around!!

PR #630, opened 2018-10-05
Jonathan Dorety

jdorety:
@keirankozlowski
Responsive-Web-Design-II project

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,14 +10,14 @@ You have experienced building fixed layout web pages, adaptive web pages, and no
 
 ## Task 2: Project Objectives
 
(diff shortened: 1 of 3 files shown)


keirankozlowski:
Jonathan Dorety
Responsive Web Design II

General:
-Your styles look nice! Only thing I'd suggest is to use a bit of margin so text isn't straight up against the edge.

HTML Index:
-See prior note about nav.

CSS Index:
-Good job setting up your break-points!
-You can use % for line-spacing.
-Your content item a width is in pixels.
-Your about services width is in pixels.
-Your vision header/headline/text p width is in pixels.
-Your text-container L-R margin is in pixels.

PR #634, opened 2018-10-05
Final commit for Fully responsive II.

JHaydenDev:
I managed to achieve all the goals and added some hover effects at various breakpoints.


JHaydenDev:
Thought I already did a pull request. 
I managed to do some stretch and added different animations at different breakpoints.

PR #641, opened 2018-10-06
Leila Berrouayel - RD II

leila100:
@keirankozlowski


keirankozlowski:
Leila Berrouayel
Responsive Web Design II

General:
-Awesome job! You animations at breakpoints look cool!
-Almost a perfect job. Keep it up!

HTML Index: 
-See prior comment about nav.
-I'd say to swap your article container tag and your div bigSec1 tag. So your container should be a div, and your bigSec1 should be an article. 

CSS Index:
-The margin for your copyright class is in flat pixels.

leila100:
Thank you.

PR #645, opened 2018-10-06
Made site fully responsive.

VonKleistman:
@keirankozlowski


keirankozlowski:
Chris Kleist
Responsive Web Design II

General:
-Nice job with your breakpoints!
-Your styles look nice.

HTML Index:
-See prior notes about semantic tag usage.

CSS Index:
-Adding max-widths on your various sub-classes isn't necessary. Using percentage based widths will make them adhere to the max-width set on the container.
-I'd recommend trying to replace your calc function stuff just for clarity's sake. It really isn't necessary for something like this.

PR #648, opened 2018-10-06
finished assignment

jrizza88:
finished assignment and have half of services page done


jrizza88:
Completed services.html as well

PR #663, opened 2018-10-06
adding code

TevvinStrong:
@keirankozlowski

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -1,1 +1,585 @@
 /* Use your CSS from RWD - I here.  IF you don't have CSS, use the solution CSS from RWD - I*/
+html,
+body,
+div,
+span,
+applet,
+object,
+iframe,
+h1,
+h2,
+h3,
+h4,
+h5,
+h6,
+p,
+blockquote,
+pre,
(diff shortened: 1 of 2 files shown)


keirankozlowski:
Tevvin Strong
Responsive Web Design II

General:
-Visually, it looks like your margin is pushing stuff off to the right.

HTML Index:
-You can simplify the two bounding divs with classes "header" and "flex". This will reduce the confusion a lot.
-Your header class should probably close after your nav.
-The div of class "photo" could be removed and put into an element selector.
-Use an address tag around your contact info.

CSS Index:
-Remove unused code prior to submission, specifically referring to the services page css.
-You lose responsiveness at 1115px; a horizonal bar forms. 
-You need to set a max-width in pixels on your container class and use percentages for everything else to do with width, including margin/padding. This is the main factor that will make things responsive.
-Your styles all look fine for the most part. Everything changes at appropriate breakpoints; you're just missing the responsiveness aspect. Just go through and do the math to turn everything to percentages and this should look fine. Please let me know if you need help and we can do some pair programming.

PR #664, opened 2018-10-06
finished

andrewogle:
all finished
@keirankozlowski


keirankozlowski:
Andrew Ogle
Responsive Web Design II

General:
-Looking nice! If you change height values back to pixels, it should fix the few overlapping elements you have in smaller views.

HTML Index:
-See prior notes about semantic tags.

CSS Index:
-If you don't need all T-B-L-R values for margin/padding, you can just use two values which equates to TB-LR.
-You can still use pixels for height and top-bottom margin/padding.
-Make sure to only re-use the elements you need for media queries. For example, if you don't need to change the margin for a certain element when moving between media queries, you just have to declare it once at the top of your cascade and it retains that value as we move between screen sizes.

PR #676, opened 2018-10-09
Chaya Otikor: Stretch

cotikor-public-plaid:


szincone:
- Fun animations and color choices
- Looks like you couldn't quite get the navbar going on the tablet version (logo over the hyperlinks), try playing with flex-direction on the .top-bar class in the header section element. (That should point you in the right direction)
- Close to pixel perfect on desktop and phone
- Great job on hitting stretch and getting your services page to be responsive. (Push that fix for the text on the buttons that you figured out end of day in stand-up)
- Maybe make more detailed and frequent commit messages.

PR #682, opened 2018-10-09
Austin Blake

dyiar:
Made desktop, tablet, and mobile view pixel good enough and fully responsive.


trevorfehrman:
Perfectly organized and readable, awesome job.  Killin it.

I've never seen list-style-type and/or position before, really cool.  Maybe you could show us that at standup if we've got time.

PR #687, opened 2018-10-09
Gina Christians RWD 2

ghost:


ashwins93:
## What's awesome?

- Site is fully responsive. The tablet view is perfect.
- Good spacing between text contents

## Tips

- Add `/* prettier-ignore */` above the lengthy css reset rule in order to stop the prettier from unwrapping it.
- The navbar in the desktop view needs more spacing between the links and the logo.
- The container class needs a `max-width` property set, currently it expands infinitely as the viewport expands.
- In the mobile view you have set the padding to `100%` on the links which is causing the content to stretch beyond the viewport. Setting horizontal padding to `100%` will result in a padding of `500px` on left and the right sides for a viewport width of `500px`. What should have been done is to set the `width: 100%` and then using `text-align: center` to match it with the design file.

PR #688, opened 2018-10-09
Eileen Cuevas

eileencuevas:
Tried to make everything pixel-close-enough


trevorfehrman:
Way above and beyond.  Excellent use of the comma selector.

Targeting elements directly might get you into hot water if you had, for example, a nav tag in your footer as well as your header which is common, but judging from the quality of this code I think you'll know to be wary of that.

You're doing really great wolf-sis

PR #689, opened 2018-10-09
Kevin Marker

kmarker1101:


css/index.css
```diff
@@ -1 +1,573 @@
...
+  border-radius: 5px;
+  text-decoration: none;
+  font-size: 1.2rem;
+  color: black;
+}
+
+/* service bottom content */
+.services .bottom-content {
+  display: flex;
+  flex-direction: row;
```
johnoro, line 356: What's the direction for (since the default is already `row`)?

johnoro: I'd also try to get the image within that `section` on point (currently, the size is looking a bit small and shrunk width-wise).

johnoro (commented):
Awesome job, especially with the formatting!

PR #690, opened 2018-10-09
Ben Tsao basic tasks completed

cbtsao47:
@ashwins93 right here~


ashwins93:
## What's awesome?

- Site looks great. It is fully responsive.
- Good use of whitespace. 
- Matches the design file perfectly.
- Comments in css.

## Tips

- Code formatting could be improved. Use a formatter like prettier to help you with that.
- Avoid transitions on `width`. Ideally you only want to animate/transition `opacity` and `transform`. Those are the properties that the browsers are optimized to animate.

PR #693, opened 2018-10-09
Daren Larson

darenlarson:


ashwins93:
## What's awesome?

- Site looks great and is fully responsive.
- Good use of whitespace
- Good job going for the stretch. Services page is fully responsive.

## Tips

- The CSS file is longer than it needs to be. Try if you can cut down some code, the media queries contains code that doesn't change across views.
- Try making more git commits.
- The cards on the services page sticks to the left edge of the screen when the view port width is about 850px, try setting the horizontal margin to `auto` and see if it fixes the UI.

PR #695, opened 2018-10-09
Daniel Mendez Responsive-Web-Design-II

dan2014:


trevorfehrman:
Lines 34-36 on index.html you can use `<br>` but this is obviously nitpicky.  Some of these fixed widths could be converted to % based widths for the sake of practice but at your max media breakpoint this isn't really needed.

Well organized, readable, controlled. A+

PR #697, opened 2018-10-09
Yasirah Krueng

ykrueng:


ashwins93:
## What's awesome?

- Site is fully responsive. Good job.
- The layout matches the design files perfectly.
- Love the animation on the home page.
- Good spacing surrounding the content.
- Awesome job implementing the expandable NavBar you are two weeks ahead of the syllabus!
- Code organization looks really neat.


### Comments on JS

- Shortest implementation of an expandable nav bar. Deserves special mention.
- `!headerNav.style.display || headerNav.style.display === 'none'` this could have been simplified to simply testing if the display property was defined and complemented by setting display to `null` when the button is toggled. Setting display property to `null` makes it fall back to css styles.

## Tips

- Changing the color to white on hover would have gone well with the blue background for the navlinks and the buttons. The nav links could use some padding - this makes a difference only when there is a background-color, like when you hover over the links.
- There is a small issue with the expandable navbar. When you are in the mobile view, you turn off the nav bar and then you expand the window, the navbar goes missing along with the hamburger icon and there is no way to bring the navbar back unless you shrink the viewport to click the icon again. This issue can be fixed by implementing the above-mentioned suggestion for the javascript.
- You are doing truly exceptional, continue helping your fellow students. I really appreciate it.

PR #698, opened 2018-10-09
Addison Stavlo - Project Complete

Addison-Stavlo:
@ashwins93


ashwins93:
## What's awesome?

- Site looks really good. It is fully responsive.
- Very good use of whitespaces. The content does not stick to the edges
- Impressive animations on nav links
- I like the background color upgrade on the nav bar
- Good code organization.

## Tips

- Try to see if you can make the background color on the navbar stretch across the page. It is a bit tricky to do that.
- Avoid animating font-weights as it alters the content dimensions. Good properties to animate are `opacity` and `transform`

PR #699, opened 2018-10-09
Tyrone Cartwright

Tyrone-Cartwright:


trevorfehrman:
Minor styling issues.  Try to add a fixed width to your logo so that it doesn't pixelate in tablet view.  Cool use of animations but I would try to focus on polish and structure before moving on to extra features.  Honestly though if it didn't take you long and it's fun then do it.

Interesting use of before and after pseudoselctors.  Just need a little more exposure to this stuff and it'll be more natural.  I'd recommend pouring over the solution code and figuring out why it works.  A lot of times it's better to reverse engineer a clean solution than cobble a bunch of fixes together.  

Good job getting everything working, you're doing awesome.

PR #701, opened 2018-10-09
Responsive Web Design II

bananaboots:
@johnoro

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -1,1 +1,405 @@
-/* Use your CSS from RWD - I here.  IF you don't have CSS, use the solution CSS from RWD - I*/
+html, body, div, span, applet, object, iframe,
+h1, h2, h3, h4, h5, h6, p, blockquote, pre,
+a, abbr, acronym, address, big, cite, code,
+del, dfn, em, img, ins, kbd, q, s, samp,
+small, strike, strong, sub, sup, tt, var,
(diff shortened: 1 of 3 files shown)


css/index.css
```diff
@@ -1 +1,405 @@
...
+h3 {
+	font-size: 2.8rem;
+}
+
+.services-btn span {
+	font-size: 1.2rem;
+}
+
+ul {
+	list-style-position:
```
johnoro, line 181: There's no property here!

css/index.css
```diff
@@ -1 +1,405 @@
...
+
+
+/* Tablet */
+
+@media only screen and (max-width: 800px) {
+
+    .container {
+		  width: 90%;
+		  max-width: 656px;
+      }
```
johnoro, line 243: There are some indentation issues; see: [prettier](https://prettier.io/)! It's really good.

css/index.css
```diff
@@ -1 +1,405 @@
...
+      .img-wrap img {
+          max-width: initial;
+      }
+      .services-header-img {
+          max-width: 656px;
+          overflow: hidden;
+      }
+      .second {
+          flex: display;
```
johnoro, line 282: I think this is backwards (should be `display: flex;`)

johnoro (commented):
Awesome work!

PR #702, opened 2018-10-09
Justin Arata: MVP Finished

jarata:


szincone:
- Good job on getting nav layout on tablet working 
- Try playing with flex-direction on the middle-bar class (see if that doesn't point you in the right direction)
- Good job on adding a border during development to all of your elements
- Good class names
- Still need desktop and phone nav bars working
- Still need main-content aligned and working

PR #703, opened 2018-10-09
Kevin Cornea

kpcornea:


ashwins93:
## What's awesome?

- Site is fully responsive. Good job.
- Good use of whitespaces.
- The layout matches the design file for desktop, tablet and the mobile.
- Good code organization.

## Tips

- On the desktop view, when the view port width is about `850px` the content starts to stick to the edges and the last link in the nav bar wraps to the next row. Think about how you would solve this issue.
- Try adding hover effects on buttons and links.

PR #707, opened 2018-10-10
Jarrad miller

jm020801:
my RWD 2


trevorfehrman:
Make more than one commit you monster.

Try to convert your font sizing and line spacing to REMs, they're really neat.
Line 167 in your css has no unit so it's silently failing.  There's a bunch of fixed dimensions that could be converted to % layout but you've converted some of them and everything works so you're demonstrating a grasp of what you're doing.  You may want to shore up a bit before sprint.

I'm really impressed with you.  You came in w/o any experience on a machine you weren't comfortable with and overcame some technical problems and managed to keep pace with an extremely rapid paced curriculum and not only that but your attitude didn't waver so you're always a pleasure to work with.

Good job wolf-bro

PR #708, opened 2018-10-10
Nedim Omerovic - Responsive Design II

Civoremo:


trevorfehrman:
Good img sizing practice, I need to look into that.  I don't think I do that and it's making me think.

Lol : margin: 3.409% 0;  That is some serious dialing in.  Gj.

Line 122 in your css has no unit.

Perfect control of specificity selection pattern.  Perfectly organized code.

You're rock solid sir.  Wp

PR #709, opened 2018-10-10
Zayna Crick

ZaynaCrick:


trevorfehrman:
Caveman approach for your negative margin is perfectly valid.  If it works it's good.

I think you could avoid needing it by setting widths for those element such that there's enough room for them both in their parent div.

I think this is also why you needed position relative on your footer cause that margin is pushing the content off the page.  My advice is to look carefully at solution code and try to understand why that works.  If you build up from working code you will sidestep a lot of problems that are huge pains to deal with.

Great job getting everything working.

PR #712, opened 2018-10-10
Kai Lovingfoss - Not completely done, but most of the issues are fixed!

tryingtokeepup:


trevorfehrman:
This looks perfect, gj boarding the struggle bus and persevering.  Having that skill is way more important than anything else.

Cool use of opacity on your mobile breakpoint. 

I think you're right on the edge of CSS feeling more natural, you already have a better grasp of it than you realize.  This code is clean.

You're doing awesome so far dude, keep it up and keeping hitting your peers up and me if you need anything clarified, you're such a pleasant person to work with.

PR #714, opened 2018-10-10
David Freitag

freitagdavid:


css/index.css
```diff
@@ -1 +1,493 @@
...
+.logo_container img{
+	height: 100%;
+	width: auto;
+}
+
+/* cta */
+
+.cta {
+
+}
```
johnoro, line 121: You have an empty selector (`.cta`); that's no good!

css/index.css
```diff
@@ -1 +1,493 @@
...
+    height: 182px;
+    margin-top: 33px;
+    background-size: contain;
+}
+
+.main-content .bottom-content {
+  margin-top: 33px;
+  width: 100%;
+  display: flex;
+	justify-content: space-between;
```
johnoro, line 194: You have some indentation inconsistencies; [prettier](https://prettier.io) is the solution!

css/index.css
```diff
@@ -1 +1,493 @@
...
+    }
+
+    .main-content .bottom-content .text-content {
+        width: 100%;
+        margin-bottom: 63px;
+        margin-top: 0;
+      }
+
+    .main-content .bottom-content .text-content:nth-child(3) {
+        margin-bottom: 0px;
```
johnoro, line 461: Remember that you can use `margin-bottom: 0;` as a shorthand over having the extra `px`.

johnoro (commented):
Awesome work; stay safe!

css/index.css
freitagdavid: Don't know how that one slipped by me.

css/index.css
freitagdavid: I've got it installed on your recommendation it's really good definitely the best beautifier I've used.

PR #727, opened 2018-11-13
RWD-II

rosilioroman:


fivegoblins:
Great job!

**Project Objectives**:

- Breakpoints introduced successfully for tablet and mobile views, with no major design issues in any views

- Fixed widths and LR margins are converted to percentages (there are a few fixed widths that you missed and left as px)

- Font sizes are converted to rem units; other responsive units are also used

**Stretch Goals**: It looks like you started building a services page but didn't complete it

**Commit history**: Great commit history; frequent commits and very descriptive

**Rating: 2**

PR #738, opened 2018-11-13
Probably messed more up than I fixed

BrannanC:
I understand the concepts, but I messed with my layout a bit too much and I introduced a handful of problems.


fivegoblins:
**Project Objectives**:

- Breakpoints successfully introduced for tablet and mobile views but there are element placement issues in these views; particularly the Innovation on Demand element.

- Fixed widths and LR margins converted to percent-based layout

- Font sizes converted to rem units

- Design generally looks 'pixel close enough' compared to design files, as instructed; except for the placement issues in tablet and mobile view mentioned above.

**Stretch Goals**: Services page is responsive, percent-based, and uses responsive units.

**Commit history**: You should be committing very frequently with descriptive commit messages that detail exactly what was completed.

**Rating: 2**

PR #764, opened 2018-11-13
Responsive Design II Cynthia

ghost:
@raylomeli


raylom:
Consider doing multiple commits an hour, with commit messages that are clear and detailed.

PR #769, opened 2018-11-13
Seriously, What Are We Naming These Things?

RaineBeauDubs:
@raylomeli


raylom:
Good job. Consider leaving multiple commits an hour with clear detailed commit messages.

PR #787, opened 2018-11-14
Commit #1

kelliraehanson:


fivegoblins:
Good work, Kelli!

**Project Objectives:**

- Breakpoints are working at 800px and 500px

- Fixed widths and LR margins converted to percentages

- Font sizes converted to rem 

- Design: you could spend some more time cleaning up the design and getting your project to more closely match the design files. Consider making the widths of some elements take up less than 100% of the container so that your text and images don't butt up against the very edges of the viewport.

**Stretch Goals**: None completed

**Rating: 2**

PR #792, opened 2018-12-13
MVP Achieved

louismagdaleno:
@gannondetroit


GannonDetroit (commented):
Hey Louis,

Great job again on this assignment and the stretch. Looks like you already acted on the recommendation I left in the last assignment and did it in this assignment regarding the commenting and white spacing on the HTML page. Fantastic to see!

I never saw a horizontal scroll bar so you mastered the lesson of using percentages to avoid such problems. Looking through your code you nailed the MVP and I don't see any thing not using REM or %'s so I'm happy you appeared to not have any issues with that. 

The only thing I can see that you might want to touch up is the nav buttons in your mobile view, They need some extra padding/space to really fit the design file.  (you also might want to fix the boxes/content in the services mobile view too. As well as the extra bit of white space at the bottom of the mobile view too.  But since this is stretch anyway it's not that big of a deal). 

overall you've done a great job my man. Very proud of how well your handling and mastering the principles of the assignments.

PR #794, opened 2018-12-14
Jack Gipson - Responsive-Web-Design II

jackgipson:
@shawnantonucci

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -65,9 +65,6 @@ h1, h2, h3, h4, h5 {
 }
 
 /* Copy and paste your work from yesterday here and start to refactor into flexbox */
-* {
-    box-sizing: border-box;
-}
 
 html {
     font-size: 62.5%;
@@ -79,17 +76,18 @@ h1, .services {
 }
 
 nav {
-	padding: 30px 0;
+	padding: 30px 0px;
     display: flex;
     width: 100%
+    
 }
 nav a {
-	margin-right: 60px;
+    /* margin-right: 6%;
+    margin-left: 2%; */
 	text-decoration: none;
 	color: #999;
 	display: flex;
-	width: 100%;
-    justify-content: center;
+    justify-content: space-between;
     font-size: 2rem;
 }
 .index {
@@ -107,11 +105,11 @@ header {
 }
 
 header a {
-    width: 50px;
+    width: 100%;
 }
 .container {
 	max-width: 880px;
-	margin: 0 auto
+	margin: 0 auto;
     padding: 10px;
 }
 .contents {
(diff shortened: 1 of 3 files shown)


css/index.css
```diff
@@ -1 +1,341 @@
...
+}
+header {
+    display: flex;
+}
+
+header a {
+    width: 50px;
+}
+.container {
+	max-width: 880px;
```
shawnantonucci: you need to set a width in the container max width is restraining to a max of that width

shawnantonucci (commented):
Try to break up your commits more

shawnantonucci:
Did you push your final changes to your project? your page is not properly acting responsively. at full screen, the whole page is stuck to the left. and when changing the size smaller your breakpoints are working but the page goes off the screen.

css/index.css
```diff
@@ -1 +1,341 @@
...
+	margin: 0 auto
```
shawnantonucci: seems like you are missing a semi colon here

css/index.css
```diff
@@ -1 +1,341 @@
...
+.top-content {
+	width: 100%;
+	display: flex;
+}
+.card {
+	margin-top: 20px;
+}
+.card2 {
+	margin-top: 20px;
+	margin-left: 70px;
```
shawnantonucci: you want to use percent here

css/index.css
```diff
@@ -1 +1,341 @@
...
+}
+.card4, .card5 {
+	margin-top: 20px;
+	margin-left: 45px;
```
shawnantonucci: same here you need percents on left or right

css/index.css
```diff
@@ -1 +1,341 @@
...
+}
+.bottom-content{
+	display: flex;
+}
+.logo {
+	margin-left: 40px;
+    display: flex;
```
shawnantonucci: remember display flex only goes on a parent container

css/index.css
```diff
@@ -1 +1,341 @@
...
+hr.line {
+	margin-top: 4%;
+	color: black;
+}
+footer p {
+    display: flex;
+    justify-content: center;
+}
+
+.btn {
```
shawnantonucci, line 128: I would clean this up

css/index.css
```diff
@@ -1 +1,341 @@
...
+	margin-top: 2%;
+}
+.bottom-portion {
+	display: flex;
+	width: 100%;
+	justify-content: space-between;
+}
+.services-info-img {
+	margin-top: 2%;
+	width: 462px;
```
shawnantonucci: you would want percent based here

css/index.css
```diff
@@ -1 +1,341 @@
...
+    .btn {
+        width: 25%;
+    }
+}
+
+/* Mobile */
+
+@media (max-width: 500px) {
+    .container {
+        padding: 0px;
```
shawnantonucci: you need to set a width here

PR #796, opened 2018-12-14
intial commit

omendez30:


css/index.css
```diff
@@ -1 +1,435 @@
...
+}
+
+.nav-bar a {
+  margin-right: 50px;
+  text-decoration: none;
+  color: gray;
+  font-size: 1.6rem;
+}
+.logo {
+  flex-direction: flex-end;
```
theryanclausen, line 181: Most flexbox styles only apply to container elements. Be sure to check your dev tools, they would have told you this attribute isn't doing anything.

css/index.css
```diff
@@ -1 +1,435 @@
...
+  flex-direction: row;
+  justify-content: space-around;
+  margin-top: 80px;
+  border-bottom: 1px solid black;
+}
+.intro-title {
+  flex-wrap: wrap;
+  align-items: center;
+  padding-top: 45px;
+  padding-left: 30%;
```
theryanclausen, line 196: these are huge paddings. They might look good at fullscreen, but the idea is to make things look good at every width. One look at the dev tools would show you that this element is overflowing your container element giving you that horizontal scroll bar.

css/index.css
```diff
@@ -1 +1,435 @@
...
+  font-size: 72px;
+  text-align: center;
+  margin-top: 30px;
+}
+.btn {
+  display: flex;
+  width: 100%;
+  justify-content: center;
+  margin-bottom: 50px;
+  padding-top: 10vh;
```
theryanclausen, line 209: be careful with mixing in vh with px. Make sure you know what your measurements are doing and be consistent at each layer. My preference is to do a max width on the page container, percentage widths and flexbox on the inner divs and pixel measured box model for stuff inside with rem based fonts.

css/index.css
```diff
@@ -1 +1,435 @@
...
+}
+.intro-title {
+  flex-wrap: wrap;
+  align-items: center;
+  padding-top: 45px;
+  padding-left: 30%;
+  padding-right: 20%;
+  width: 95%;
+  height: 30vh;
+  font-size: 72px;
```
theryanclausen, line 200: use rem

css/index.css
```diff
@@ -1 +1,435 @@
...
+/* vvvvvvv SERVICES AND OTHER FEATURE vvvvvv */
+.section-one {
+  display: flex;
+  flex-direction: row;
+  margin: 50px 0px;
+  font-size: 1.6rem;
+  font-weight: bold;
+}
+.middle-img {
+  width: 900px;
```
theryanclausen, line 238: try something more responsive with this so it scales with your window size.

css/index.css
```diff
@@ -1 +1,435 @@
...
+}
+/*^^^^^^^^^ TITLE AND IMG ^^^^^^^^^*/
+
+/* vvvvvvv SERVICES AND OTHER FEATURE vvvvvv */
+.section-one {
+  display: flex;
+  flex-direction: row;
+  margin: 50px 0px;
+  font-size: 1.6rem;
+  font-weight: bold;
```
theryanclausen, line 235: your p text up top is bold and the p text below is regular. Be consistent.

theryanclausen (commented):
developer tools
I think you need to start using your developer tools a little more. Open up that sidebar with 'ctrl shift i' and hit that button by the number 1 then hover over and see how your spacing is being applied. It's color coded like the picture at number 5, blue is content, green is padding, yellow is border, orange is margin. I think the only thing holding you back right now is you are losing track of the space you can't see with the margins and padding. Use those dev tools and they are going to help you out in a big way. Try to get this thing 100% responsive with no scroll bar at a width wider than 400px. That should be the first thing you work on this break. Keep working hard, you are making good progress.

PR #797, opened 2018-12-14
Aislynn Edmiston- Responsive-Web-Design-II

AislynnEdmiston:


css/index.css
```diff
@@ -65,143 +65,236 @@ h1, h2, h3, h4, h5 {
...
 
+ .links {
+    display: flex;
+    justify-content: space-between;
+    color: #b0b0b0;
+    /* border: 2px solid red; */
+}
+
+.links a {
+    padding-right: 4rem;
```
jeffyjkang, line 99: this is always going to create that horizontal scroll bar, when you do a padding, why dont you try making the parent container flex and doing space between or around, whatever you see fit

jeffyjkang (commented):
keep up the good work

PR #798, opened 2018-12-14
First and only pull

GregoryUrban:
@theryanclause


css/index.css
```diff
@@ -1 +1,286 @@
...
+.titlepic {
+  size: 30%;
+}
+
+section .title-img {
+  /* border: 1px solid black; */
+}
+
+.titlepic {
+width: 50%;
```
theryanclausen: here's a quick hint: the percentage measurement applies to the property of it's parent element. You have this image inside a div that is taking up roughly half the page, so you are telling this image to be half the width of half the page.

theryanclausen (commented):
Great work so far!

css/index.css
GregoryUrban: Thanks Ryan - I refactored all the non-font sizes to %, and it cascaded into sizes much better

css/index.css
```diff
@@ -1 +1,270 @@
...
+header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+}
+
+nav {
+  width: 70%;
+  display: flex;
+  justify-content: inherit;
```
theryanclausen, line 98: that is a nifty trick

css/index.css
```diff
@@ -1 +1,270 @@
...
+  width: 50%;
+  display: flex;
+  flex-wrap: wrap;
+  flex-direction: column;
+  align-items: center;
+}
+
+section .title button {
+  width: 40%;
+  background-color: white;
```
theryanclausen, line 128: buttons have some annoying built in styling. If you want it to look more like the design file, throw in a border-style none and then add your own border after. Remember the cascade, your border must come after you take away the default. As far as the use of button here, since this isn't a form and you don't need any of the button form functionality, I would probably just style an a tag to look like a button. That way you are starting from scratch.

css/index.css
```diff
@@ -1 +1,270 @@
...
+  color: gray;
+}
+
+section {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-between;
+  align-items: center;
+  border-bottom: 2px solid black;
+  padding: 11% 0 12% 0;
```
theryanclausen: this is too much space. I would avoid using responsive measurements on anything on the vertical axis. That would include height properties and top and bottom padding. Just because you know responsive measurements, doesn't mean you always have to use them.

css/index.css
```diff
@@ -1 +1,270 @@
...
+    flex-direction: column;
+    border-bottom: 1px solid gray;
+ 
+  }
+
+  nav a {
+    border-top: 1px solid gray;
+    width: 100%;
+    text-align: center;
+    height: 10rem;
```
theryanclausen, line 231: notice what happens when you get your window down under 400px. The links start drifting up. Try to avoid the height property whenever possible. Pixel measured padding is usually your best option for height issues. You are better off setting distances from your content with padding and margin than setting a height and hopping all of your content fits in properly.

theryanclausen (commented):
It is awesome to see the amount of progress you made this week. I think you have found a way to wrap your head around this stuff. You have picked up some really cool tricks along the way. Be careful about using heights. You can almost always get around that using box model and padding. Measure out from your content, don't set a height and hope your content fits inside. As far as the responsive stuff goes, I think your best bet is going to be max width on the page container, percentages on your larger elements inside the container, pixel measurement box model for elements inside of that, rems for text. Just because a person needs bigger letters, doesn't mean they also need bigger margins. Keep working hard and the next time you struggle with something, just remember what a difference a week can make.

PR #800, opened 2018-12-14
Cam Perry - RWDII 

cpscott1:
@shawnantonucci

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -1,1 +1,361 @@
-/* Use your CSS from RWD - I here.  IF you don't have CSS, use the solution CSS from RWD - I*/
+/* http://meyerweb.com/eric/tools/css/reset/
+   v2.0 | 20110126
+   License: none (public domain)
+*/
+
+html, body, div, span, applet, object, iframe,
+h1, h2, h3, h4, h5, h6, p, blockquote, pre,
+a, abbr, acronym, address, big, cite, code,
+del, dfn, em, img, ins, kbd, q, s, samp,
+small, strike, strong, sub, sup, tt, var,
+b, u, i, center,
+dl, dt, dd, ol, ul, li,
+fieldset, form, label, legend,
+table, caption, tbody, tfoot, thead, tr, th, td,
+article, aside, canvas, details, embed,
+figure, figcaption, footer, header, hgroup,
+menu, nav, output, ruby, section, summary,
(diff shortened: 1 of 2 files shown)


shawnantonucci:
Try to commit more within what you are working on. For example, instead of doing added HTML after adding all the HTML I would instead do:  "Added navbar to index.html", "created a profile view", "added in the address in the footer"

css/index.css
```diff
@@ -1 +1,376 @@
...
+
+header nav {
+  display: flex;
+  width: 100%;
+  vertical-align: top;
+  margin-top: 35px;
+}
+
+header nav a {
+  margin-right: 65px;
```
shawnantonucci, line 103: To be fully responsive your width needs to be percent based. try using percents on the margin

shawnantonucci (commented):
Everything looks really good here Cam. My only suggestion i have for you is to try adding in comment to separate your sections so you understand what's going on and when other go look at your code

css/index.css
```diff
@@ -1 +1,376 @@
...
+.item3 p {
+	padding-bottom: 20px;
+}
+
+.item4 {
+	border: 1px solid black;
+	height: 20px;
+	line-height: 20px;
+	margin: 20px;
+	padding: 20px;
```
shawnantonucci, line 291:
```suggestion
	padding: 20px 5% 20px 5%;
```

shawnantonucci: when using a single parameter for padding it adds 40px on all sides and does use percentage base. I would do something like this

css/index.css
```diff
@@ -1 +1,376 @@
...
+.item4 p {
+	padding-bottom: 20px;
+}
+
+.item5 {
+	border: 1px solid black;
+	height: 20px;
+	line-height: 20px;
+	margin: 20px;
+	padding: 20px;
```
shawnantonucci, line 303: same on the margin and padding here

shawnantonucci:
taking a look at your page again you are using borders to view your elements Great job! Before you push your work make sure to delete or comment out the borders so they are not visible after. Also, it looks like you Innovation on demand and button are scrolling off the screen when resizing. Your nav links on the mobile view are bunched together and missing the lines separating them. I would also Change the flex direction of the features, about, etc sections to column so that it looks cleaner and more compact.

PR #802, opened 2018-12-14
Nicholas.Rafeek/Responsive-Web-Design-II

Reyaad510:


css/index.css
```diff
@@ -1 +1,369 @@
...
+    display: flex;
+    align-items: center;
+    
+}
+
+
+header nav ul .item {
+    flex: 1;
+}
+
```
JoeBugajski, line 121: Interesting choice, going with the 'flex: 1'. One of my favorite parts of being a PM is seeing each student's unique approach to solving problems. I honestly wasn't sure what this property did when I first saw it here, so I had to go educate myself on w3schools! I'll have to remember that one.

css/index.css
```diff
@@ -1 +1,369 @@
...
+         border: 1px solid black;
+         font-size: 1.6rem;
+         background: white;
+         width: 45%;
+         height: 32px;
+         margin-top: 10px;
+         cursor: pointer;
+         transition: 250ms ease-in-out;
+     }
+
```
JoeBugajski, line 174: I like the button animation! People are usually super impressed by even simple animation. So it's worthwhile to have a few tricks like this up your sleeve. Keep playing around with it!

JoeBugajski (commented):
Your pages look really solid. You definitely went above and beyond, hitting several stretch goals. It seems you've quickly mastered media queries with no problem. It's really not as crazy as it might have seemed at first. 

- Step one: resize the page

- Step two: add styles until page looks right at the new size.

Not that hard, right?

PR #803, opened 2018-12-14
Brandon Gardner

BrandonGardner2:


css/index.css
```diff
@@ -1 +1,498 @@
...
+
+@media (max-width: 500px) {
+  .container {
+    margin: 20px auto;
+  }
+
+  /*Nav Area*/
+  .nav {
+    padding-right: 6%;
+    width: 100vw;
```
theryanclausen, line 421: I think the only thing giving you a horizontal scroll bar is these vw properties. The difference between vw and percentage width is percentage width respects the container. You have a little bit of overflow at the mobile media break but other than that, this thing is flawless.

theryanclausen (commented):
You continue to crank out good work. You are reaching out when you need help and stretching when you can. If you keep that up, you will continue to do well here. Great job

PR #805, opened 2018-12-14
proj

bcuz:


style.css
```diff
@@ -0,0 +1,262 @@
...
+	margin-bottom: 20px;
+}
+
+.fact-item ul li {
+  margin-bottom: 15px;
+}
+
+/* tablet */
+@media (min-width: 500px) {
+	.container {
```
jeffyjkang, line 180: you may want some margin on the mobile version of the services page, if you check the edges of the content are so close to the width that it almost looks cut off

jeffyjkang (commented):
great job with mvp and stretching with the services page

PR #809, opened 2018-12-14
basic HTML and CSS pages

max-keller427:


jeffyjkang:
Hey Max great job, i still see some horizontal scroll, but you can fix it in two areas, you may want to work on the nav, it creates a horizontal, and the transitioning of the middle image right before it goes into the tablet view, it busts out of the side, see if you can fix this, take a look at how that elements are nested and try to figure out a way to do this.

PR #810, opened 2018-12-14
Kristea Gamel

bykristea:
RWD II initial commit to set up html and css

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -48,7 +48,12 @@ table {
 
 /* Set every element's box-sizing to border-box */
 * {
-    box-sizing: border-box;
+  box-sizing: border-box;
+}
+
+html {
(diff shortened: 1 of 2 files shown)


css/index.css
```diff
@@ -1 +1,380 @@
...
+
+/* //////////////////////////////// */
+@media all and (max-width: 800px) {
+   .container {
+       width: 100%;
+       padding: 20px;
+   }
+
+   header {
+       margin-top: 32px;
```
theryanclausen, line 244: You may have noticed that you can see content above your fixed nav when you scroll down. Remember that background color only applies to elements inside of the border.

theryanclausen (commented):
Amazing work here. You have done a great job putting everything we have learned together to create an awesome webpage. Keep up the good work.

PR #812, opened 2018-12-14
Laurhens - RWD II

LDinho:


css/index.css
```diff
@@ -1 +1,393 @@
...
+    align-items: center;
+    justify-content: flex-end;
+    height: 100%;
+    width: 30%;
+}
+
+header {
+    display: flex;
+    justify-content: space-around;
+    height: 455px;
```
theryanclausen, line 122: be careful with fixed heights, especially now that we are using rems. If somebody cranks up the text size you are going to have stuff spilling out all over the place. It is better to measure out from your content with padding and margin than it is to set a height and hope your content always fits inside.

css/index.css
```diff
@@ -1 +1,393 @@
...
+    margin-bottom: 30px;
+    margin-top: 136px;
+}
+
+/*.header-container {*/
+/*!* TODO update to flexbox!*/
+/*}*/
+
+.header-wrapper {
+    position: relative;
```
theryanclausen, line 133: this is an odd choice. You doing a lot of extra work double wrapping this content and using absolute positioning and everything when all you have to do is remember your cross axis on the header flex and you are good to go.

css/index.css
```diff
@@ -1 +1,393 @@
...
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+    text-align: center;
+    width: 480px;
```
theryanclausen, line 138: this is not a responsive width. This is giving you a horizontal scroll bar.

css/index.css
```diff
@@ -1 +1,393 @@
...
+        width: auto;
+    }
+}
+
+/*mobile view media query*/
+
+@media(max-width: 500px) {
+
+    nav {
+        height: 372px;
```
theryanclausen, line 334: remember what I said about setting heights. Look at what happens when you change the browser settings to extra large font

css/index.css
```diff
@@ -1 +1,393 @@
...
+        padding: unset;
+        width: 105%;
+    }
+
+    .nav-item-wrapper a {
+        display: flex;
+        justify-content: center;
+        align-items: center;
+        border: 1px solid;
+        height: 54px;
```
theryanclausen, line 359: box model. Also, you are getting a 2px thick border between the links because you are doing your border all the way around the links. If you look closely at the design file you will see that there are no side borders on these links. If you chose a top or bottom border property you should be able to keep a consistent border width.

css/index.css
```diff
@@ -1 +1,393 @@
...
+    .header-h1 {
+        font-size: 8.6rem;
+    }
+
+    .nav-item-wrapper {
+        display: flex;
+        flex-direction: column;
+        font-size: 2rem;
+        padding: unset;
+        width: 105%;
```
theryanclausen, line 351: be careful with widths over 100%. It's like negative margins. If you are using negative margins, there is probably some extra space you are not accounting for. If you are using >100% measurements, your container is probably too small. In this case however, your container is the right size. I'm not sure if you have this here to make the side borders disappear, but whatever the reason, this and the fixed width on your header text is giving you a horizontal scroll bar.

theryanclausen (commented):
Good job overall. I think you were just overthinking that header element a little too much. You must remember the cross axis with your flexbox. It's good you have the skills to out think yourself, but you should look for the simple solutions before you do something crazy with absolute positioning. Remember fixed width measurements are not responsive and setting height to elements are generally not a good idea. Flexbox can usually get you out of having a set height and will protect you from overflow. Unintentional overflow can be a page killer. I would suggest you take a look at flexbox zombies over the break to really hammer home all of that knowledge. Do not be afraid to reach out to the help channel or me directly if you find yourself in an overthinking situation. Keep up the good work. See if you can get rid of the horizontal scroll bar.

PR #813, opened 2018-12-14
Tara Sherman

tarateel:
Initial commit

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -84,6 +84,7 @@ h2 {
     display: flex;
     justify-content: space-between;
     flex-wrap: wrap;
+    /* border: 0.5px solid; */
 }
 
 nav {
@@ -140,6 +141,7 @@ header .getting-started .start {
     flex-direction: row;
     justify-content: center;
     flex-wrap: nowrap;
+    border: 1px solid black;
 }
 
 header .snippet {
@@ -157,6 +159,13 @@ main {
     flex-wrap: wrap;
 }
 
+main .top {
+    width: 100%;
+    padding-top: 4%;
+    display: flex;
+    justify-content: space-between;
+}
+
 main .top h2 {
     margin-bottom: 4.3%;
 }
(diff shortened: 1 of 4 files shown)


css/index.css
```diff
@@ -1 +1,559 @@
...
+    align-items: center;
+}
+
+nav .nav-items {
+    width: 70%;
+	font-size: 1.85rem;
+	justify-content: space-between;
+}
+
+.nav a:hover, a:visited, a:link, a:active {
```
theryanclausen: this is a little bit of over kill. Editing the text decoration and color of nav a is enough to override the defaults for all of these states.

css/index.css
```diff
@@ -1 +1,559 @@
...
+    .container {
+        width: 100vw;
+        display: flex;
+        flex-direction: column;
+        flex-wrap: wrap;
+        align-items: center;
+    }
+
+    nav {
+        width: 98vw;
```
theryanclausen: this is where the trouble starts. vw and vh are pretty tricky to work with and you have to be very careful that your added margins don't push you beyond the size of your page container. I usually reserve vw and vh for instances when I need one element to take over the entire window. Like how on facebook when you look at a picture, it tints the page you were on and loads the picture and comments in the center of the screen. vw does not respect the size of the container it is in. I think your best move to stay responsive is to use max width on your page container and percentages on the divs inside your container. Set one standard and subdivide from there.

theryanclausen (commented):
Good job on this. I can see how hard you worked on this. I think you went a little overboard on the responsive measurements in some instances. My rule of thumb is: max width on the container element, percentages and flexbox for the larger divs, pixel measured box model for everything inside, rem for text. Just because a person may need larger fonts doesn't mean they need larger margins. This is just the way I lay my pages out. There isn't necessarily a right or wrong way so you can continue to work with vws if you can make it work. Just remember that one of the goals of fluid responsiveness is to eliminate the horizontal scroll bar at all window sizes. Once you get down around 300px, it's a bit over kill, but 350-400px is a good target to shoot for. Keep up the good work.

PR #814, opened 2018-12-14
Steve Kutz

stevekutz:


css/index.css
```diff
@@ -1 +1,377 @@
...
+/* Set every element's box-sizing to border-box */
+* {
+  box-sizing: border-box;
+}
+
+html{
+ /*  border: 4px solid purple; */
+  font-family: 'Titillium Web', sans-serif;
+  display: flex;
+  max-width: 880px;
```
theryanclausen, line 59: be careful styling html and body elements. They have some odd behaviors in general, but especially with older browsers. The container div has pretty much been universally adopted.

css/index.css
```diff
@@ -1 +1,377 @@
...
+  flex-direction: column;
+ /* max-width: 800px; */
+  margin: 0 auto;
+  height: 100%;
+  font-size: 62.5%;
+  background: palegoldenrod;
```
theryanclausen: Love that color

css/index.css
```diff
@@ -1 +1,377 @@
...
+  justify-content: space-between;
+  align-items: center;
+  padding: 60px 0;
+  margin-top: .9rem;
+}
+
+.banner {
+  border: 1px solid brown;
+  display: flex;
+  height: 360px;
```
theryanclausen, line 133: setting heights will get you in trouble. It is better to measure out from your content with padding and margin than setting a height and hoping everything fits in. Overflow can be a page buster.

css/index.css
```diff
@@ -1 +1,377 @@
...
+  .bottom .bottom-content {
+    width: 30%;
+  }
+
+
+}
+
+
+@media (max-width: 500px) {
+
```
theryanclausen, line 308: looks like you kind of ran out of steam down here or you haven't pushed the latest. I assume this is still a work in progress given all the guide borders.

theryanclausen (commented):
Really well done getting pixel close enough at desktop and tablet, mobile still needs work. Keep working hard and reaching out when you need help.

PR #815, opened 2018-12-14
Refactored index for rem and %.  Adjusted and refined for near pixel perfect.

joestricker:
Completed MVP.


gsamaniego41 (commented):
Really impressive work Joe. Site is fully responsive. Smooth transitions. Multiple pages built. Only thing I would recommend is to inspect the middle image in index.html and header image in services.html. They stretch vertically when resizing window. It could be the properties of the container they're in. Let me know if you want to work on it together.

PR #816, opened 2018-12-14
initial commit. + Font-size to rem

NathanWaddell121107:


css/index.css
```diff
@@ -1 +1,373 @@
...
+        flex-direction: column;
+        margin: 0;
+        padding: 0;
+    }    
+
+    @keyframes example {
+        from {background-color: blue;}
+        to {background-color: lightgray;}
+      }
+
```
JoeBugajski, line 277: Nice animation! I'm glad you went for a stretch goal. I would encourage you to keep digging into animation. People tend to be really impressed by even simple animation. It's a nice skill to have under your belt to show clients that you have "magic powers."

PR #819, opened 2018-12-14
Amanda Thomas

Memitaru:


css/index.css
```diff
@@ -1 +1,306 @@
...
+* {
+    box-sizing: border-box;
+}
+
+html, body {
+    height: 100%;
+    font-family: 'Titillium Web', sans-serif;
+}
+
+html {
```
lolax, line 60: Spot-on :)

css/index.css
```diff
@@ -1 +1,306 @@
...
+    font-size: 1.6rem;
+}
+
+header nav a:hover {
+    text-decoration: underline;
+    color: black;
+}
+
+header .logo {
+    margin: 28px 0 0 91px;
```
lolax, line 105: Try to make right & left margins (and padding) percentages for greater responsiveness. It might look a little less pixel perfect, but will be more dynamic.

css/index.css
```diff
@@ -1 +1,306 @@
...
+    color: white;
+}
+
+/* Main Content Styles */
+
+.main-content {
+   border-top: 2px solid black; 
+   border-bottom: 2px solid black;
+   margin: 75px 0 35px 0;
+   padding: 4% 0;
```
lolax, line 148: Avoid using percentages for top & bottom margin or padding; usually screen width is what we accommodate since vertical scroll bars are perfectly fine, especially on a phone.

css/index.css
```diff
@@ -1 +1,306 @@
...
+    .cta img {
+        display: none;
+    }
+
+    .main-content .middle-img {
+       width: 100%;
+    }
+
+    .main-content {
+       margin: 75px 3%px 35px;
```
lolax, line 226: Looks like you accidentally have both a percentage & px here, which might be causing a small disruption.

lolax (commented):
Great work on this project, Amanda! You've demonstrated an excellent understanding of responsive design.

PR #823, opened 2018-12-14
John O'Keefe - Responsive-Web-Design-II

jokeefe1:
Rewrote entire project from scratch using only flex for practice. I was tempted to use CSS grid for the navigation container to reorder the logo but decided to use flex-direction: row-reverse instead.

I feel I am making steady improvement and my code is becoming more concise with each round of practice. I look forward to your feedback. 

J.

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -76,11 +76,13 @@ html {
 
 h1 {
     font-size: 2.4rem;
+    font-family: var(--Bangers);
 }
 
 h3 {
     font-size: 2.2rem;
-    font-weight: 600;
+    font-family: var(--Bangers);
+
 }
 
 p, a, address {
@@ -91,6 +93,10 @@ p, a, address {
     width: 95%;
     max-width: 880px;
     margin: 0 auto;
+}
+
+.horizontal-line {
+    color: var(--dark);
 }
 
 /* Custom styles
@@ -141,8 +147,7 @@ nav {
 }
 
 .big-text {
-    font-size: 5rem;
-    font-family: var(--Bangers);
+    font-size: 7rem;
 }
 
 .header-img {
@@ -190,6 +195,7 @@ nav {
 
 .contact {
     padding: 2rem;
+    max-width: 25rem;
 }
 
 .address {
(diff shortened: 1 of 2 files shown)


css/index.css
```diff
@@ -1 +1,306 @@
...
+	border-spacing: 0;
+}
+
+/* Default styles
+============================ */
+
+:root {
+
+    /* Site colors */
+    --dark: #111;
```
lolax, line 56: Excellent use of variables to update themes en masse.

css/index.css
```diff
@@ -1 +1,306 @@
...
+
+html {
+    font-size: 62.5%;
+}
+
+body {
+    font-family: var(--primaryFont);
+}
+
+h1 {
```
lolax, line 82: I like that you're reassigning rem-based font sizes for these elements, after their styles were cleared in the reset.

css/index.css
```diff
@@ -1 +1,306 @@
...
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    padding: 2rem;
+}
+
+.nav-link {
+    width: 100%;
+    text-decoration: none;
+    border-top: 0.1rem solid var(--lightAccent);
```
lolax, line 124: I don't think I've seen rems used for border thickness, but it seems perfectly reasonable to me.

css/index.css
```diff
@@ -1 +1,306 @@
...
+.title {
+    font-family: var(--secondaryFont);
+}
+
+/* Middle Section
+============================ */
+
+.middle-img {
+    width: 100%;
+    padding: 0 1%;
```
lolax, line 190: I'm curious to know why you switch between percentages and rems for padding. I do appreciate the use of rem instead of px for padding & margins - I imagine it would make things look much better for users with special font settings.

css/index.css
```diff
@@ -1 +1,306 @@
...
+    /* Header */
+
+    .header {
+        flex-direction: row;
+        justify-content: space-between;
+    }
+
+    .header-img {
+        display: block;
+        width: 100%;
```
lolax, line 303: Is there a reason I'm not seeing that your max-width should be smaller than your width here?

lolax (commented):
Excellent work, John! Glad to see that you challenged yourself to translate from grid to flex. I appreciate your variables and liberal use of rem 💯

PR #825, opened 2018-12-15
Levi-Huynh initial draft for RD II project

Levi-Huynh:


css/index.css
```diff
@@ -1 +1,1153 @@
...
+	width: 100%;
+	display: flex;
+	flex-direction: row;
+	flex-wrap: wrap;
+	align-content: center;
+	justify-content: space-between;
+	line-height: 36px;  
+}
+
+@keyframes increase{
```
lolax, line 107: Cool animations!

css/index.css
```diff
@@ -1 +1,1153 @@
...
+	display: flex;
+	margin-left: 0;
+	animation: increase 2s ease-in; 
+	animation-iteration-count: 5;
+	animation-direction: alternate;
+	}
+
+div nav a{
+	/* border: 1px solid green; */
+	margin: 20px 5% 20px 5%;
```
lolax, line 135: You could shorted this to: 
`margin: 20px 5%;`

css/index.css
```diff
@@ -1 +1,1153 @@
...
+}
+
+.copyright{
+	/* border: 1px solid purple; */
+	display: flex;
+}
+
+.SVcontainer{
+	/* border: 1px solid red; */
+	width: 800px;
```
lolax, line 533: Try to use percentages for `width` properties. However, you can use px for `max-width`.

lolax (commented):
Great work with responsive design concept, Levi! You used percentages and rems consistently and with great precision. I would aim to write more concise CSS, which I imagine will be made easier with tools like mixins and nesting in LESS. Keep it up! 💯

PR #828, opened 2018-12-15
Responsive-Web-Design-II

Dionuta:


GannonDetroit (commented):
Hey Tay,

so something happened between this project and your last one. 

You site seems to be set to the window size, not a container because everything is getting messed up when I expand and shrink it, this is really harming your styling and the UI experience, you also have horizontal scroll bars which transforming everything from pixel to percentages should have eliminated. 

go into your css and re-evaluate the several instances where you still have pixels on aspects that effect width. Also double check your use of what's in your body and .max properties as well. It might explain why your website is getting harmed when I play with it's expansion/shrinking in the window. Use Dev tools in chrome to help you out with this. 

let me know if you need more help with anything.

PR #831, opened 2018-12-15
Responsive-Web-Design-II MVP

willediger:
@gannondetroit

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -83,7 +83,7 @@ a {
 }
 
 .container {
-  width: 880px;
+  max-width: 880px;
   min-height: 1px;
   margin: 0 auto;
 }
@@ -98,7 +98,7 @@ nav {
   vertical-align: top;
   margin-top: 13px;
   display: flex;
-  width: 606px;
+  width: 68.86%;
   justify-content: space-between;
 }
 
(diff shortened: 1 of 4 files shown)


GannonDetroit (commented):
Will, you have a bit of scroll bar action on your desktop version that appears when I get close to the breakpoint of your tablet view. The problem is most likely that you still have some pixel uses in your CSS (mainly for margins), I recommend using dev tools to figure out the percent number to convert those pixels to. Pixels create horizontal scroll bars, percents don't. (I only noticed this on the desktop view, your tablet and mobile views appeared solid.).

your transitions on the stretch is amazing. It looks like some professional Squarespace level stuff. Great job on that!!! 

(ignore previous review, that was meant for your Responsive Design I project).

willediger:
@GannonDetroit I fixed the horizontal scrollbar issue on the desktop view & lower widths. It wasn't caused by pixel margins, but by not having a max-width on the middle image. I set that to 100%, and changed the margin-left on the image to padding-left so that the width + padding of the image would have a max-width of 100%, rather than just the width of the image having a max-width 100% (which would still cause a horizontal scroll bar since the margin isn't taken into account by the max-width).

PR #837, opened 2018-12-15
Miguel Diaz - Responsive-Web-Design-II

diaz4674:
@ielvisd finished main objectives for project.


css/services.css
```diff
@@ -0,0 +1,243 @@
...
+font-family: 'Bangers', cursive;
+letter-spacing: 1px;
+margin-bottom: 15px;
+}
+
+/* Your code starts here! */
+
+html {
+   max-width: 100%;
+   font-size: 62.5%;
```
ielvisd: Good use of font-size % to standardize your rem units.

ielvisd, line 70: Good use of font size % to standardize rem units.

css/index.css
```diff
@@ -1 +1,515 @@
...
+    margin-top: 5rem;
+}
+.container {
+    font-size: 62.5%;
+    margin: 0 auto;
+    max-width: 880px;
+}
+
+
+@media(max-width: 800px) {
```
ielvisd, line 222: You could make your code more dry by only adding the lines that changed to your media query.

css/index.css
```diff
@@ -1 +1,515 @@
...
+
+.copyright {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    margin-bottom: 10rem;
+    margin-top: 5rem;
+}
+.container {
+    font-size: 62.5%;
```
ielvisd, line 216: Good use of max width and `margin : 0 auto` to center your site.

ielvisd (commented):
Good job, you'll want to add some padding to parts of your paragraphs that are right up against the images or the edges of the site. Also I like the fixed nav but think of how you might re-arrange it so that it doesn't take up the majority of the real estate in mobile view.

PR #841, opened 2018-12-15
jacquettabommuluri-Responsive-Web-Design-II

jacquetta:
…ated README.md checklist.
@jeffyjkang


jeffyjkang (commented):
great work Jackie, even got the services page up on this project as well, I can still see a horizontal scroll bar on your transitions, this is most likely due to the fact that some of your children element/content is busting out the side, see if you can find where they are, maybe you can try to set the width/ max-width to 100% of the parent container, and/or reduce the size altogether when they hit certain breakpoints. awesome work

PR #842, opened 2018-12-15
Response II Added HTML/CSS from prior project.  Initial commit.

blubrandi:
@jeffyjkang


jeffyjkang (commented):
nicely done on the mvp, only slight thing i would correct is in the mobile view, font-size: 10rem, creates a horizontal scroll bar due to the font busting out of the container, but small fix

PR #843, opened 2018-12-15
Fully refactored Responsive Great Idea Website

jthicks91:
@GannonDetroit

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -1,1 +1,290 @@
-/* Use your CSS from RWD - I here.  IF you don't have CSS, use the solution CSS from RWD - I*/
+/* http://meyerweb.com/eric/tools/css/reset/ 
+   v2.0 | 20110126
+   License: none (public domain)
+*/
+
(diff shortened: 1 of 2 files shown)


GannonDetroit:
Hey Jordan,

Looks like you started to do a bit of what I mentioned in the last assignment regarding adding some commenting in the HTML file. Good job and keep doing it, do more if you can. 

your site looks great. The percentages you put in have removed the horizontal scroll bar and makes the transition very seamless and smooth. 

The only exception is when you transition from desktop to tablet view, when you get super close to the tablet breakpoint, for a brief moment, a horizontal scroll bar appears. The reason for this is that there is a part in your code that you missed refactoring it from px to percentage. (it's an element in/near the nav, just a hint). 

that minor detail aside, you 've shown mastery over this topic. You know how to use REM's and percentages and add the breakpoints and why it all works. Very proud to see how well you've come on the subject matter. Keep working through this break and stay on top of things. 

you keep doing this and you have an awesome future ahead of you.

PR #850, opened 2018-12-15
Responsive Redesign

ThatGuyUKnow593:
Redesigned using percents for layout, and used rem for text size for a smooth and responsive look

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -91,7 +91,7 @@ header .navLink {
 }
 
 header a {
-	width: 4rem;
+	width: 10%;
 	margin: auto;
 	color: gray;
 	font-size: 1.5rem;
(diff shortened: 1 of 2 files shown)


GannonDetroit:
Trent, your break points are very smooth and I see you have the viewport meta tag in this one, good job!

The only thing I can find to point out is that you have horizontal scroll bars on your tablet and mobile views which you shouldn't have in this project since it's supposed to be in all percentages. Based on what I'm seeing in your code I'd recommend you take a close look at how the various "max-width" you have effect things. You have more than you most likely need and is causing problems because the scroll bar is mainly showing extra white space on the right hand side of the page. So you have everything right, just need to debug and figure out where that extra white space is coming from (dev tools and boarders can help with debugging). The readme also tells you to use 880px as a max width, make sure you follow that the readme asks for because that's essentially the requirements a boss would expect you to execute on in a job so you can't miss things like that. 

I'm glad this one has 2 commits but you still are going to want to add more to your projects (doesn't need to be a ton, but at least a few).

PR #851, opened 2018-12-15
Mohamed Khaloti - Responsive Web Design II Challenge

Khalotis11:


gsamaniego41 (commented):
Really great job Moe! Site is fully responsive all the way down to a 500px screen.

PR #856, opened 2018-12-15
Shota Kikozashvili -mvp

ShotaKiko:
@GannonDetroit

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -73,7 +73,7 @@ html{
     font-size: 1.6rem;
     max-width: 880px;
     margin: 0 auto;
-    border: 2px dashed red;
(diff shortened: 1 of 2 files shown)


GannonDetroit (commented):
Ok so all the issues I listed in the Responsive web design I seemed to be handled in this one lol. 

Great job in making sure there was no horizontal scroll bar, everything looks great and just like the design file. (only nic-picky exception is that the font size of the nav bar is a bit small, but not that big of a deal right now). 

Great job again on the commit history, commenting in your code and use of the border technique. 

I really don't have much feedback to give because you truly knocked this out of the park and did a great job!. Congratulations :)

PR #859, opened 2018-12-15
responsive design II

claridiva2000:


css/index.css
```diff
@@ -1 +1,280 @@
...
+  max-width:880px;
+  margin: auto;
+}
+
+nav,header,#section1,#middle,.section2,.container4, address,footer{
+margin: auto;
+width: 95%;
+font-size: 2em;
+}
+
```
JoeBugajski, line 79: There's a small, tiny bug where the layout breaks while viewport > 801px && viewport < 838px. It's because you have width at 95% on the header. Maybe you should try using a bit of margin to keep content from the edges, rather than 95% width.
```suggestion
 width: 100%;
```
Actually, after playing around with it, it seems width: 100% doesn't really have an effect that I can see. You might not need a width property set on the header at all. It's just that 95% that's causing a problem. 

Also, try using classes for most things, as opposed to Id's. Because Id's have more weight than classes, it can be difficult to style a page that is full of Id's.

JoeBugajski (commented):
Project looks very good overall. Over the break, see if you can finish the services page, just for fun.

PR #861, opened 2018-12-15
MVP finished, working on animations

douglasjordan2:


css/index.css
```diff
@@ -1 +1,566 @@
...
+		width: 92.5vw;
+		margin: 0 auto;
+	}
+
+	.copyright {
+		padding-top: 5%;
+	}
+
+	/* Services Styles */
+
```
JoeBugajski, line 396: Are these services styles just an artifact from your previous project? Because I'm not seeing a services html file.

douglasjordan2:
Yep, those are just artifacts, I forgot to remove those when I submitted within the time limit. I plan on redoing the whole project in LESS to get familiar with it. I’ll resubmit when I finish that up.

PR #863, opened 2018-12-15
responsive webdesign 2/incomplete

RodBromberg:
@ielvisd

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -1,1 +1,426 @@
-/* Use your CSS from RWD - I here.  IF you don't have CSS, use the solution CSS from RWD - I*/
+/* http://meyerweb.com/eric/tools/css/reset/ 
+v2.0 | 20110126
+License: none (public domain)
+*/
+
+html,
+body,
+div,
+span,
+applet,
+object,
+iframe,
+h1,
+h2,
+h3,
+h4,
+h5,
+h6,
+p,
+blockquote,
+pre,
+a,
+abbr,
+acronym,
+address,
+big,
+cite,
+code,
+del,
+dfn,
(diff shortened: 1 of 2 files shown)


css/index.css
```diff
@@ -1 +1,426 @@
...
+    letter-spacing: 1px;
+    margin-bottom: 15px;
+}
+
+.container {
+    width: 880px;
+    align-items: center;
+    justify-content: center;
+    margin: 0 auto;
+}
```
ielvisd, line 168: max-width instead of width would get rid of your horizontal scroll bar. 
Good use of `margin: 0 auto` to center your content.

css/index.css
```diff
@@ -1 +1,426 @@
...
+
+* {
+    box-sizing: border-box;
+}
+
+html,
+body {
+    height: 100%;
+    font-family: 'Titillium Web', sans-serif;
+    font-size: 62.5%;
```
ielvisd, line 150: Good use of 62.5% font size to standardize your rem sizes across your site.

css/index.css
```diff
@@ -1 +1,426 @@
...
+.a-logo {
+    order: 0;
+    width: 25%;
+    text-align: center
+}
+
+
+/* Media Query Tablet */
+
+@media only screen and (max-width: 800px) {
```
ielvisd, line 301: Good use of media queries to change the look of your site.

ielvisd (commented):
Look into what flexbox properties you can use to center the title on tablet/mobile views. Also you could add a rem font size to your headers and footer text to make those not tiny. Good job overall.

PR #870, opened 2018-12-18
@jeffyjkang, Responsive Web Design II

cfelps:


jeffyjkang (commented):
very nicely done mvp chris, due to the fact that you have flex end on the great idea logo, when you transition from the desktop to tablet view, the logo gets cut-off by the vertical scoll bar, see if you can find a way where it will stay inside, and your specs look slightly off from the design files but you completed the fully responsive aspect of it and achieved no horizontal scroll which is the most important thing. good job

PR #871, opened 2018-12-19
Init pull

Nathan-Culver:


GannonDetroit (commented):
Nathan, everything looks good with the design file, but your tablet view seems to have a good amount of extra white space under the "get started" button and the line beneth it, see if you can close that gap a little. The Mobile view for "innovation on demand" is a bit off as well. It's too small. I have a horizontal scroll bar when i use it. This is most likely because of the pixel units you're still using. Speaking of, you're still using pixels for quite a few things in the CSS. Try to use your dev tools to turn those pixels into percertages. (padding, margins, things like that).

PR #872, opened 2018-12-23
Re-forked and re-pulled repo!

millan-figueroa:
@shawnantonucci

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,21 +1,139 @@
-<!doctype html>
+<!--Millan Figueroa -- Responsive Design II: Fully Responsive // HTML File-->
+<!DOCTYPE html>
 
 <html lang="en">
-<head>
-  <meta charset="utf-8">
-  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <head>
+        <meta charset="utf-8" />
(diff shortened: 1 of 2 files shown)


shawnantonucci:
Awesome job on your project Millan. My only suggestions for you are on your navbar when you hover over a link it pushes everything away on the page. One way to solve this is to add a set height on the nav container. This will fix the rest of the page moving but you will notice the links now move. You need to set a width/height of the container element so that when it animates it has room within it container to grow without pushing the next element. Also, I would set the main container width to be the size of the design file so that it doesn't take up the full page width even if you are viewing it from a 4k monitor or anything greater than 1200

index.html
```diff
@@ -1,21 +1,139 @@
...
+<!--Millan Figueroa -- Responsive Design II: Fully Responsive // HTML File-->
+<!DOCTYPE html>
 
 <html lang="en">
-<head>
-  <meta charset="utf-8">
-  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <head>
+        <meta charset="utf-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
```
shawnantonucci, line 7: Everything in the header could be bunched together you don't really need whitespace here it looks cleaner without it

PR #875, opened 2019-01-08
Jonathan Gistand - Responsive-Web-Design-II

JonathanGis:


css/index.css
```diff
@@ -1 +1,306 @@
...
+	font-size: 1.8rem;
+	line-height: 1.25;
+}
+
+.middle-img {
+    display: flex;
+    justify-content: space-around;
+    align-content: column;
+	margin: 0 auto;
+	padding: 2% 0%;
```
lolax, line 149: Try to avoid using percentages for vertical (top & bottom) padding or margins because screen size is generally dictated by its width

css/index.css
```diff
@@ -1 +1,306 @@
...
+}
+
+
+.contact-header {
+    display: flex;
+    justify-content: center;
+    align-content: flex-start;
+    padding-top: 12px;
+    margin: 4rem;
+	padding-right: 750px;
```
lolax, line 181: On the converse, try to use percentages for left-right padding/margins so that they scale with screen size :)

css/index.css
```diff
@@ -1 +1,306 @@
...
+    align-content: flex-start;
+    padding-top: 16px;
+    margin: 24px;
+	padding-right: 675px;
+	font-size: 1.8rem;
+	line-height: 1.25;
+}
+
+.footer {
+	display: flex;
```
lolax, line 198: It's a technical necessity and in this case it's a small difference, but I encourage you to try to use consistent spacing to make your code easier for you and others to read.

css/index.css
```diff
@@ -1 +1,306 @@
...
+	.getstarted-button {
+		display: flex;
+		justify-content: flex-end;
+		padding-right: 370px;
+}
+
+
+/* Mobile Specs */
+
+@media (max-width: 500px) {
```
lolax, line 248: Nice work on both your breakpoints!

lolax (commented):
Really nice work on this project, Jonathan! I left a few comments pointing towards potential improvements below. You've demonstrated a solid understanding of responsive design. Keep it up! 💯

PR #876, opened 2019-01-09
Responsive-Design-2

Christian-Ford:


css/index.css
```diff
@@ -1 +1,330 @@
...
+	width: 100%;
+	size: .5rem;
+}
+
+.navContainer {
+	display: flex;
+	flex-direction: row;
+	justify-content: center;
+	width: 70%;
+	height: 8%;
```
lolax, line 97: Avoid using percentages for heights and top/bottom margins & padding. Vertical scroll bars are perfectly fine and we don't want content to get super squished if someone's window height is small, they will just have to scroll more downwards.

css/index.css
```diff
@@ -1 +1,330 @@
...
+
+}
+
+h1 {
+	font-size: 7.5rem;
+}
+
+.button {
+	border: 1px solid black;
+	width: 30%;
```
lolax, line 150: Nice job with the percentages for widths

css/index.css
```diff
@@ -1 +1,330 @@
...
+  }
+
+  .circle {
+    display: none;
+  }
+
+}
+/* =========== Mobile View  ===========*/
+
+@media (max-width: 500px) {
```
lolax, line 248: Good job with your breakpoints 💯

lolax (commented):
Great work on this project, Christian. You were thorough with your switch to a fully responsive design. My only recommendation is that you avoid using percentages for height and top/bottom padding & margins because we don't want to shrink everything on small screen dimension, just make it more minimal and longer/narrower. Vertical scrolling is expected so it's okay for elements to be a little tall, horizontal scrolling is what we want to eliminate. Nice work! 💯

